PR #14, opened 2017-11-14
Improvements

urbanserj:
* Speedup dcos-net startup time
* Reduce number of threads from 123 to 18
* Fix erldns error: `[error] Failed to load zones: enoent`
* Set dump_log_write_threshold to 10

We don't need 100 async threads and 100000 open files since we don't have so much i/o operations.

### DIFF
--- a/config/sys.config
+++ b/config/sys.config
@@ -4,7 +4,8 @@
     ]},
 
     {mnesia, [
-        {dir, "/var/lib/dcos/navstar/mnesia"}
+        {dir, "/var/lib/dcos/navstar/mnesia"},
+        {dump_log_write_threshold, 10}
     ]},
 
     {dcos_rest, [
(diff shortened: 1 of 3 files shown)


config/sys.config
```diff
@@ -65,6 +65,7 @@
     ]},
 
     {kernel, [
+        {error_logger, silent},
```
GoelDeepak: Would it also take away messages that could help in debugging?

apps/dcos_dns/src/dcos_dns_router.erl
```diff
@@ -58,17 +58,21 @@ default_resolvers() ->
     application:get_env(?APP, upstream_resolvers, Defaults).
 
 %% @private
--spec(find_upstream(Name :: binary(), Labels :: [binary()]) -> [upstream()]).
-find_upstream(_Name, [<<"mesos">>|_]) ->
+-spec(find_upstream(Labels :: [binary()]) -> [upstream()]).
```
GoelDeepak: I am not really sure what this change is buying us

PR #21, opened 2017-12-01
Handle new TASK_STARTING status

urbanserj:
https://jira.mesosphere.com/browse/DCOS_OSS-1924


apps/dcos_l4lb/testdata/state6.json
```diff
@@ -0,0 +1,249 @@
+{
```
GoelDeepak: looks like this is a handcrafted state.json.  Can we take a real state.json instead?

PR #22, opened 2017-12-06
Get rid of erldns udp and tcp servers

urbanserj:
https://jira.mesosphere.com/browse/DCOS_OSS-1946

### DIFF
--- a/apps/dcos_dns/src/dcos_dns_router.erl
+++ b/apps/dcos_dns/src/dcos_dns_router.erl
@@ -13,7 +13,7 @@
 -spec(upstreams_from_questions(dns:questions()) -> ordsets:ordset(upstream())).
 upstreams_from_questions([#dns_query{name=Name}]) ->
     Labels = dcos_dns_app:parse_upstream_name(Name),
-    Upstreams = find_upstream(Name, Labels),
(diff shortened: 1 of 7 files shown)


apps/dcos_dns/src/dcos_dns_router.erl
```diff
@@ -10,43 +10,19 @@
...
+-spec(upstreams_from_questions(dns:questions()) -> [upstream()] | erldns).
 upstreams_from_questions([#dns_query{name=Name}]) ->
     Labels = dcos_dns_app:parse_upstream_name(Name),
-    Upstreams = find_upstream(Labels),
-    lists:map(fun validate_upstream/1, Upstreams);
-
-%% There is more than one question. This is beyond our capabilities at the moment
+    find_upstream(Labels);
 upstreams_from_questions([Question|Others]) ->
+    %% There is more than one question. This is beyond our capabilities at the moment
```
GoelDeepak: It will be good to log a message here. Maybe at debug level?

urbanserj: I've added debug logging [here](https://github.com/dcos/dcos-net/pull/22/files#diff-284042ed3d64edbe734c34310f0555e4R21).

PR #28, opened 2017-12-19
Add enable_ipv6

urbanserj:
https://jira.mesosphere.com/browse/DCOS_OSS-1993

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
@@ -45,7 +45,7 @@
 
 -record(vip_be2, {
     protocol = erlang:error() :: protocol(),
-    vip_ip = erlang:error() :: inet:ip_address(),
+    vip_ip = erlang:error() :: inet:ip_address() | {name, {binary(), binary()}},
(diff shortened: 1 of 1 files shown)


apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
```diff
@@ -223,6 +223,24 @@ flatten_vips(VIPDict) ->
         ),
     ordsets:from_list(VIPBEs).
 
+-spec(maybe_disable_ipv6([vip_be2()]) -> [vip_be2()]).
+maybe_disable_ipv6(VIPBEs) ->
+    case application:get_env(dcos_l4lb, enable_ipv6, true) of
```
GoelDeepak: I am under impression that IPv6 is failing on all latest centos kernels. If that is the case then no dcos cluster will come up with latest centos. Maybe till we find the cause (and fix) we could turn ipv6 off by default

urbanserj: I'll disable it by default in dcos PR [1]

[1] https://github.com/dcos/dcos/pull/2232/files

PR #29, opened 2017-12-26
Downgrade OTP to 20.0.5

asridharan:
Due to the `hostname` check validation errors for x509 certificates downgrading OTP version to 20.0.5.

---
This change is [Reviewable](https://reviewable.io/reviews/dcos/dcos-net/29)

### DIFF
--- a/apps/dcos_dns/src/dcos_dns_app.erl
+++ b/apps/dcos_dns/src/dcos_dns_app.erl
@@ -31,7 +31,6 @@ maybe_start_dcos_dns(false) ->
 maybe_start_dcos_dns(true) ->
     Ret = dcos_dns_sup:start_link(true),
     maybe_start_tcp_listener(),
-    maybe_start_http_listener(),
     Ret.
 
 stop(_State) ->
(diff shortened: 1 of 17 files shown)


urbanserj:
We don’t need to downgrade anything in dcos-net repository, we should downgrade only erlang package in dcos/dcos repo.

asridharan:
@urbanserj the ref in `dcos/dcos` repo points to the head of `dcos-net` at this point. However, there are quite a few commits between the head and the commit introducing the OTP upgrade. We need all those commits. Hence, explicitly reverted the upgrade while keeping all the commits.

asridharan:
Ah !! I misunderstood. Didn't realize we have an explicit package for OTP in DC/OS:
https://github.com/dcos/dcos/blob/master/packages/erlang/buildinfo.json

Will look at just downgrading the OTP package.

gabrielhartmann:
Is there another PR I should be watching?

asridharan:
@gabrielhartmann will close this out. @urbanserj was able to reproduce the problem and should have a fix shortly. Will update the JIRA with the PRs.

gabrielhartmann:
Cool, thanks.

PR #32, opened 2017-12-31
[WIP] Docker build

asridharan:
Added a Dockerfile to build of an alpine container on Mac OSX and Linux.

---
This change is [Reviewable](https://reviewable.io/reviews/dcos/dcos-net/32)


Dockerfile
```diff
@@ -0,0 +1,63 @@
+FROM alpine:3.6
```
urbanserj: Could you please use semi-official `erlang:20.2.2-alpine` image?

How can I use it to build and test dcos-net? I guess we need to add some instructions.

Right now I use dcos-docker to build, test, and run dcos-net. Please see [1] [2]

[1] https://github.com/dcos/dcos-net#development
[2] https://github.com/dcos/dcos-net/blob/master/bin/dev.sh

asridharan: @urbanserj looked at `bin/dev.sh`. You still need to mount your code as volume in `dcos-docker` right? Also, given that we are compiling within a DC/OS cluster how would the other `dcos-net` react (running on the agents) to this new `dcos-net`. Was hoping to have a stand-alone dev environment.

urbanserj: We use this image for dcos-net dependencies, e.g. [mesos_state](https://github.com/dcos/mesos_state/blob/master/.circleci/config.yml#L6) or [dcos-netapi](https://github.com/mesosphere/dcos-netapi/blob/master/.circleci/config.yml#L6). I'd like to use it for dcos-net's CI build, but CircleCI runs docker in non-privileged mode (without `ip_vs`). In dcos we use vanilla version of OTP, so I'm fine to use semi-official erlang image for unit tests and cover all major features in dcos integration tests as well.

BTW I will take a look at the failed test case.

asridharan: I would prefer trying to capture the OTP dependency in our CI within this repo rather than relying on DC/OS integration tests to catch it (fail early). Would it be ok, if we have a `mesosphere/erlang-otp` image with the OTP version we compile against and base all our images of that OTP version? Given that we anyway compile our OTP versions with DC/OS thought that might be appropriate?

urbanserj: @asridharan dcos-docker mount your home directory by default or you can add the following line to `make-config.mk`:

```
CUSTOM_VOLUMES := -v /Users/youruser:/Users/youruser:rw
```

dcos supports migration from the latest stable version, so anyway your new dcos-net build has to work with the old ones or you can run that build on each node if you run several `bin/dev.sh` simultaneously.

asridharan: True. Though you have to explicitly opt in for this when you run `autoconf` in `dcos-docker`.  That said, wouldn't you still need to mount your Github repo into `dcos-docker` since the `bin/dev.sh` script assumes you are running at the root of your source tree:
https://github.com/dcos/dcos-net/blob/master/bin/dev.sh#L5

That said I think we should have both options available, of working with `dcos-docker` (which is pretty important for CI and integration testing) as well working standalone which is important for development velocity.

urbanserj: https://github.com/asridharan/dcos-net/pull/1

asridharan: Will update the docker build to use Erlang/OTP: 2.2

urbanserj:
See https://github.com/dcos/dcos-net/pull/40

asridharan:
Thanks @urbanserj going to close this one in favor of #40 .

PR #34, opened 2018-02-12
Add support for VIPs on UCR bridge network

urbanserj:
https://jira.mesosphere.com/browse/DCOS_OSS-2070

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
@@ -38,10 +38,12 @@
 }).
 
 %% API
--export([start_link/0]).
--export([push_vips/1,
-         push_netns/2]).
-
+-export([
+    start_link/0,
+    push_vips/1,
+    push_netns/2,
+    local_port_mappings/1
+]).
 
 %% gen_statem behaviour
 -export([init/1, terminate/3, code_change/4, callback_mode/0]).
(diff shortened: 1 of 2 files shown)


apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -176,7 +200,8 @@ push_routes(Namespace, #state{routes = Routes, route_mgr = RouteMgr}) ->
 push_services(Namespace, State = #state{last_configured_vips = VIPs}) ->
     lists:foreach(fun({VIP, BEs}) -> add_service(VIP, BEs, Namespace, State) end, VIPs).
 
-do_reconcile(VIPs0, State0 = #state{ns = Namespaces}) ->
+do_reconcile(VIPs, State0 = #state{ns = Namespaces}) ->
+    VIPs0 = vips_port_mappings(VIPs),
     VIPs1 = process_reachability(VIPs0, State0),
     Routes = ordsets:from_list([VIP || {{_Proto, VIP, _Port}, _Backends} <- VIPs1]),
     State1 = State0#state{last_received_vips = VIPs0, last_configured_vips = VIPs1, routes = Routes},
```
GoelDeepak: VIPs instead of VIPs0

urbanserj: we must store original value into the state for future reconciliation, because port mappings can be changed, but lashup wouldn't get such changes.

urbanserj:
@GoelDeepak I've enabled test_vip for UCR containers on bridge mode, I guess it's enough and it should cover all possible cases.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
GoelDeepak: yeah, this make sense

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -272,7 +298,7 @@ maintain(VIPs0, State = #state{ns = Namespaces, routes = Routes, last_configured
                     do_update_routes(NewRoutes, Routes, Namespace, State),
                     apply_diff(Diff, Namespace, State)
                   end, Namespaces),
-    State#state{last_configured_vips = VIPs1, last_received_vips = VIPs0, routes = NewRoutes}.
+    State#state{last_configured_vips = VIPs1, last_received_vips = VIPs, routes = NewRoutes}.
```
GoelDeepak: Shouldn't this be VIPs0 as well based on the same logic as with reconcile?

PR #40, opened 2018-03-26
Docker build

urbanserj:
https://jira.mesosphere.com/browse/DCOS_OSS-2309

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ VERSION         ?= $(shell git describe --tags)
 BASE_DIR         = $(shell pwd)
 ERLANG_BIN       = $(shell dirname $(shell which erl))
 REBAR            = $(shell pwd)/rebar3
+DOCKER_RUN       = $(shell pwd)/bin/docker.sh
 
 .PHONY: rel deps test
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@ PACKAGE         ?= dcos-net
 VERSION         ?= $(shell git describe --tags)
 BASE_DIR         = $(shell pwd)
 ERLANG_BIN       = $(shell dirname $(shell which erl))
-REBAR            = $(shell pwd)/rebar3
+REBAR            = $(shell pwd)/bin/run.sh ./rebar3
```
asridharan: @urbanserj suggestion:
Instead of having a `run.sh`, can we just have `docker-image` and `docker` as a target.

`make docker-image` can build the docker image and `make docker` can do what `run.sh` does?

This will make it simple for developers to build without docker in case they don't wish to build and run within a docker container?

urbanserj: I've added `docker`, `docker-image` and `docker-*` targets. So now you can run any makefile targets in docker, e.g. `make docker-compile docker-ct docker-dialyzer`. Could you please review this PR again?

Makefile
```diff
@@ -56,3 +57,16 @@ rel:
...
 	${REBAR} release -d
+
+##
+## Docker
+##
+
+docker-image:
+	@${DOCKER_RUN} true
+
+docker-%:
```
asridharan: This is cool 👍

PR #56, opened 2018-06-20
Always depend on TaskState, it is more accurate

urbanserj:
Mesos Operator API doesn't send any updates for TaskStatus if
client already receives TaskState == TASK_RUNNING. There is no
syncronization betweenn TaskState and TaskStatus events. However,
task networking information is always correct if agent reports
that task is running.

Link: https://github.com/apache/mesos/blob/d418f15/include/mesos/master/master.proto#L601-L606
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-3539


apps/dcos_net/src/dcos_net_mesos_state.erl
```diff
@@ -202,7 +202,9 @@ handle_task_updated(Obj, State) ->
     Task = mget(<<"status">>, Obj),
     FrameworkId = mget(<<"framework_id">>, Obj),
     Task0 = mput(<<"framework_id">>, FrameworkId, Task),
-    handle_task(Task0, State).
+    TaskState = mget(<<"state">>, Obj),
+    Task1 = mput(<<"state">>, TaskState, Task0),
```
GoelDeepak: Should we put a comment that we are overwriting the status.state here?

PR #63, opened 2018-07-30
Skip restablish vtep

mainred:
Reconfiging mac address of an interface will reset its static ip neighbours.
Thus, static ip neighbours of vtep interface will be removed,
if dcos_overlay is restarted, which results in traffic break.

Fix this by ignoring reconfiging an existing vtep interface.

Close-Bug: [DCOS_OSS-3855](https://jira.mesosphere.com/browse/DCOS_OSS-3855)


apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -171,20 +171,76 @@ maybe_create_vtep(_Overlay = #mesos_state_agentoverlayinfo{backend = Backend},
...
-    case {application:get_env(dcos_overlay, enable_ipv6, true), VTEPIP6} of
+    case vxlan_interface_present(Pid, VXLan) of
+        false ->
+            dcos_overlay_netlink:iplink_add(Pid, VTEPNameStr, "vxlan", VNI, 64000),
+            {ok, _} = dcos_overlay_netlink:iplink_set(Pid, ParsedVTEPMAC, VTEPNameStr);
+        {true, match} ->
+            lager:debug("Attributes of ~p is up-to-date.", [VTEPNameStr]);
+        {true, notmatch} ->
+            lager:debug("Attributes of ~p are not update-to-date, and will be recreated.", [VTEPNameStr]),
+            dcos_overlay_netlink:iplink_delete(Pid, VTEPNameStr),
```
GoelDeepak: Removing an existing vtep1024 interface might break customer setting if the customer is using this interface for some other purpose and happens to install DC/OS. We would like to fail explicitly instead of silently changing customer setting. 

We should just check if vtep1024 is already present and configure only if it is not present.

apps/dcos_overlay/src/dcos_overlay_netlink.erl
```diff
@@ -7,7 +7,7 @@
 %% Application callbacks
 -export([start_link/0, stop/1, ipneigh_replace/5, iproute_replace/6, bridge_fdb_replace/4,
         iplink_show/2, iplink_add/5, iplink_set/3, iprule_show/2,
-        iprule_add/5, make_iprule/4, match_iprules/2, is_iprule_present/2,
+        iprule_add/5, make_iprule/4, match_iprules/2, is_iprule_present/2, iplink_show_interface/2, iplink_delete/2,
```
GoelDeepak: `interface` is redundant in `iplink_show_interface`

apps/dcos_overlay/src/dcos_overlay_netlink.erl
```diff
@@ -80,6 +80,22 @@ iplink_show(Pid, Ifname) ->
...
   netlink_request(Pid, getlink, [], Link).
 
+%% eg. iplink_delete(Pid, "vtep1024") ->
+%%         {ok,[]}
+iplink_delete(Pid, Ifname) ->
+  Attr = [{ifname, Ifname}, {ext_mask, 1}],
+  Link = {packet, arphrd_netrom, 0, [], [], Attr},
+  netlink_request(Pid, dellink, [], Link).
+
+iplink_show_interface(Pid, Ifname) ->
```
GoelDeepak: This function could be simplified based on the below comment.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -171,20 +171,76 @@ maybe_create_vtep(_Overlay = #mesos_state_agentoverlayinfo{backend = Backend},
...
+    case dcos_overlay_netlink:iplink_show_interface(Pid, VTEPNameStr) of
+        {ok, DeviceLinkInfo} ->
+            Match = vxlan_attributes_match(DeviceLinkInfo, VXLan),
+            {true, Match};
+        {error, ErrorCode, ResponseMsg} ->
+            lager:debug("Failed to find ~p for error_code: ~p, msg: ~p", [VTEPNameStr, ErrorCode, ResponseMsg]),
+            false
     end.
 
+vxlan_attributes_match([{address, Address} |Right], VXLan) ->
```
GoelDeepak: we could get rid of this function all together

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -171,20 +171,76 @@ maybe_create_vtep(_Overlay = #mesos_state_agentoverlayinfo{backend = Backend},
...
       _ ->
-          ok = try_enable_ipv6(VTEPNameStr),
-          {ParsedVTEPIP6, PrefixLen6} = parse_subnet(VTEPIP6),
-          {ok, _} = dcos_overlay_netlink:ipaddr_replace(Pid, inet6, ParsedVTEPIP6, PrefixLen6, VTEPNameStr)
+          ok = try_enable_ipv6(IfName),
+          {ParsedVTEPIP6, PrefixLen6} = parse_subnet(IPAddressV6),
+          {ok, _} = dcos_overlay_netlink:ipaddr_replace(Pid, inet6, ParsedVTEPIP6, PrefixLen6, IfName)
+    end.
+
+vxlan_interface_present(Pid, VXLan) ->
```
GoelDeepak: This function would need a re-org based on earlier comment

apps/dcos_overlay/src/dcos_overlay_poller.erl
mainred: If we have a recognizable name for vtep, e.g. dcos.1024, we won't be afraid of removing consumer-defined interfaces. And all these code works fun, and should remain.

mainred:
recheck

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -171,19 +173,75 @@ maybe_create_vtep(_Overlay = #mesos_state_agentoverlayinfo{backend = Backend},
...
+    #mesos_state_vxlaninfo{vtep_name = VTEPName} = VXLan,
+    VTEPNameStr = binary_to_list(VTEPName),
+    case dcos_overlay_netlink:iplink_show(Pid, VTEPNameStr) of
+        {ok, ?LINK_INFO(LinkInfo)} ->
+            Match = vxlan_attributes_match(LinkInfo, VXLan),
+            {true, Match};
+        {error, ErrorCode, ResponseMsg} ->
+            lager:error("Failed to find ~p for error_code: ~p, msg: ~p", [VTEPNameStr, ErrorCode, ResponseMsg]),
+            false;
+        {ok, []} ->
```
urbanserj: Could you please explain me this case? Why does it return `{true, match}` here?

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -171,19 +173,75 @@ maybe_create_vtep(_Overlay = #mesos_state_agentoverlayinfo{backend = Backend},
     VTEPNameStr = binary_to_list(VTEPName),
     ParsedVTEPMAC = list_to_tuple(parse_vtep_mac(VTEPMAC)),
     {ParsedVTEPIP, PrefixLen} = parse_subnet(VTEPIP),
-    dcos_overlay_netlink:iplink_add(Pid, VTEPNameStr, "vxlan", VNI, 64000),
-    {ok, _} = dcos_overlay_netlink:iplink_set(Pid, ParsedVTEPMAC, VTEPNameStr),
-    {ok, _} = dcos_overlay_netlink:ipaddr_replace(Pid, inet, ParsedVTEPIP, PrefixLen, VTEPNameStr),
-    case {application:get_env(dcos_overlay, enable_ipv6, true), VTEPIP6} of
+    case vtep_present(Pid, VXLan) of
```
urbanserj: Could you please move this new code to new function `create_vtep/2`?

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -171,19 +173,75 @@ maybe_create_vtep(_Overlay = #mesos_state_agentoverlayinfo{backend = Backend},
...
+          ok = try_enable_ipv6(IfName),
+          {ParsedVTEPIP6, PrefixLen6} = parse_subnet(IPAddressV6),
+          {ok, _} = dcos_overlay_netlink:ipaddr_replace(Pid, inet6, ParsedVTEPIP6, PrefixLen6, IfName)
+    end.
+
+vtep_present(Pid, VXLan) ->
+    #mesos_state_vxlaninfo{vtep_name = VTEPName} = VXLan,
+    VTEPNameStr = binary_to_list(VTEPName),
+    case dcos_overlay_netlink:iplink_show(Pid, VTEPNameStr) of
+        {ok, ?LINK_INFO(LinkInfo)} ->
```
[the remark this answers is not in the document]
mainred: Yes, It's better to use a structured record.

apps/dcos_overlay/src/dcos_overlay_poller.erl
mainred: In fact, {ok, []} will never occur in normal circustance except that we are running dcos-net with non-root user, as in ource circle test with `circleci` user.
```
netlink_request(Pid, Type, Flags, Msg) ->
  Uid = list_to_integer(string:strip(os:cmd("id -u"), right, $\n)),
  case Uid of
    0 -> %% root
      gen_netlink_client:rtnl_request(Pid, Type, Flags, Msg);
    _ ->
      %% here here
      io:format("Would run fun ~p with flags ~p and argument ~p~n", [Type, Flags, Msg]),
      {ok, []}
  end.
```
I can add comment on that for explanation, but it's better to support non-root user to run dcos-net for security.

apps/dcos_overlay/src/dcos_overlay_poller.erl
mainred: add vtep link and set mac address can be moved to `create_vtep/2`, but it's better to keep  `vtep_present` here, because it judge where or not should ( maybe ) we create vtep, right?

apps/dcos_overlay/src/dcos_overlay_poller.erl
urbanserj: It's not "for security". I'd prefer to remove [1] and disable all overlay tests if non-root user starts them. That's what we are doing in l4lb common tests [2].

@GoelDeepak what would you say?

[1] https://github.com/dcos/dcos-net/blob/08ca22f/apps/dcos_overlay/src/dcos_overlay_netlink.erl#L176-L194
[2] https://github.com/dcos/dcos-net/blob/ce621b7/apps/dcos_l4lb/test/dcos_l4lb_netlink_SUITE.erl#L41

mainred:
@urbanserj, PTAL

apps/dcos_overlay/src/dcos_overlay_poller.erl
urbanserj: could you please move this `case` expression to new function `maybe_create_vtep_link/2`, rename `create_vtep/2` to `create_vtep_link/2`, and rename `config_l3/2` to `create_vtep_addr/2`?

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -158,21 +159,47 @@ config_overlay(Overlay, State) ->
...
+    } = VXLan,
+    VTEPNameStr = binary_to_list(VTEPName),
+    case vtep_present(Pid, VXLan) of
+        false ->
+            lager:debug("~p will be created.", [VTEPNameStr]),
+            create_vtep(VXLan, State);
+        {true, true} ->
+            lager:debug("Attributes of ~p is up-to-date.", [VTEPNameStr]);
+        {true, false} ->
+            lager:info("Attributes of ~p are not update-to-date, and vtep will be recreated.", [VTEPNameStr]),
```
urbanserj: could you please change log level to `debug`? we will add a better logging here in `master` branch (it's one of the features dcos-net in dc/os 1.12)

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -158,21 +159,47 @@ config_overlay(Overlay, State) ->
...
+            dcos_overlay_netlink:iplink_delete(Pid, VTEPNameStr),
+            create_vtep(VXLan, State)
+    end,
+
+    config_l3(VXLan, State).
+
+config_l3(VXLan, #state{netlink = Pid}) ->
```
urbanserj: `s/config_l3/create_vtep_addr/`

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -185,6 +212,37 @@ maybe_create_vtep(_Overlay = #mesos_state_agentoverlayinfo{backend = Backend},
           {ok, _} = dcos_overlay_netlink:ipaddr_replace(Pid, inet6, ParsedVTEPIP6, PrefixLen6, VTEPNameStr)
     end.
 
+vtep_present(Pid, VXLan) ->
```
urbanserj: nit: `s/vtep_present/check_vtep_link/`

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -185,6 +212,37 @@ maybe_create_vtep(_Overlay = #mesos_state_agentoverlayinfo{backend = Backend},
...
+        {ok, [#rtnetlink{type=newlink, msg=Msg}]} ->
+            {unspec, arphrd_ether, _, _, _, LinkInfo} = Msg,
+            Match = vxlan_match(VXLan, LinkInfo),
+            {true, Match};
+        {error, ErrorCode, ResponseMsg} ->
+            lager:info("Failed to find ~p for error_code: ~p, msg: ~p", [VTEPNameStr, ErrorCode, ResponseMsg]),
+            false
+    end.
+
+vxlan_match(VXLan, Link) ->
```
urbanserj: nit: `s/vxlan_match/match_vtep_link/`

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -302,4 +360,82 @@ deserialize_overlay_test() ->
     Msg = mesos_state_overlay_pb:decode_msg(OverlayData, mesos_state_agentinfo),
     ?assertEqual(<<"10.0.0.160:5051">>, Msg#mesos_state_agentinfo.ip).
 
+vxlan_match_test() ->
+    VNI = 1024,
+    VTEPIP = <<"44.128.0.1/20">>,
+    VTEPMAC = <<"70:b3:d5:80:00:01">>,
+    VTEPMAC2 = <<"70:b3:d5:80:00:02">>,
+    VTEPNAME = <<"vtep1024">>,
+    FAKE_LINK_INFO = [
```
urbanserj: that's very long data: we have a style checker that has 30-line limit for functions, but unfortunately it doesn't force the rule for `TEST` code. I'd recommend to move this data to `apps/dcos_l4lb/test/vtep_link.data` and read that file here using `file:consult/1`.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -158,21 +159,47 @@ config_overlay(Overlay, State) ->
     maybe_create_vtep(Overlay, State),
     maybe_add_ip_rule(Overlay, State).
 
+create_vtep(VXLan, #state{netlink = Pid}) ->
```
urbanserj: nit: could you please move this function below and rename it to `create_vtep_link`? I believe it would be easier to understand that code if you use the following order:

```
maybe_create_vtep() ->
maybe_create_vtep_link() ->
match_vtep_link() ->
match() ->
create_vtep_link() ->
create_vtep_addr() ->
```

apps/dcos_overlay/src/dcos_overlay_poller.erl
urbanserj: @GoelDeepak is it still the case?

PR #64, opened 2018-08-01
disables erldns cache to improve latency in resolving the DNS query

GoelDeepak:
It was observed that when a new task is launched then it would
take close to 5 seconds to resolve the DNS even though erldns
had been updated with the new record. This was happening because
of the erldns cache. This patch fixes the issue by disabling the
erldns cache. This might have an impact on erldns but we believe
it should be minimal as all the queries are local to an agent.

### DIFF
--- a/apps/dcos_dns/include/dcos_dns.hrl
+++ b/apps/dcos_dns/include/dcos_dns.hrl
@@ -13,6 +13,7 @@
 -define(DCOS_DIRECTORY(Prefix), <<Prefix, ".thisdcos.directory">>).
 -define(DCOS_DOMAIN, ?DCOS_DIRECTORY("dcos")).
 -define(MESOS_DOMAIN, ?DCOS_DIRECTORY("mesos")).
+-define(DCOS_DNS_TTL, 5).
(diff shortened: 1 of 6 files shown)


apps/dcos_dns/test/dcos_dns_mesos_tests.erl
```diff
@@ -53,6 +53,11 @@ hello_overlay_test_() ->
         {"hello-host-vip-0-server", fun hello_overlay_host_vip/0}
     ]}.
 
+dns_cache_test_() ->
```
urbanserj: `dcos_dns_mesos_tests.erl` contains only tests for mesos state listener. Could you please move that new test to `dcos_dns_SUITE.erl`? Also I believe you even don't need lashup to cover this case.

GoelDeepak: I kept it there because it required minimal changes.  I will try to move it to `dcos_dns_SUITE.erl`. lashup is just a wrapper around erldns in those test.

PR #66, opened 2018-08-03
Add support for OTP 21

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-3655


urbanserj:
@GoelDeepak request for review

apps/dcos_l4lb/test/dcos_l4lb_lashup_vip_listener_SUITE.erl
```diff
@@ -70,6 +70,7 @@ lookup_failure3(Config) ->
 
 lookup_vip(_Config) ->
   {ok, _} = update_de8b9dc86_marathon(),
+  timer:sleep(100),
```
GoelDeepak: does it that long?

urbanserj: Well, it's hard to say. We don't have metric for this. It's a flaky test. What value do you think we should use here?

GoelDeepak: Sorry, I was under impression that the input to `sleep` is in seconds. Looks like it's in milliseconds. We should be fine then

PR #68, opened 2018-08-08
Make mesos listener error tolerant

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-3930

### DIFF
--- a/apps/dcos_net/src/dcos_net_mesos_listener.erl
+++ b/apps/dcos_net/src/dcos_net_mesos_listener.erl
@@ -458,19 +458,25 @@ handle_task_port_mappings(<<"DOCKER">>, TaskObj) ->
 
 -spec(handle_port_mappings(jiffy:object()) -> [task_port()]).
 handle_port_mappings(PortMappings) when is_list(PortMappings) ->
-    lists:map(fun handle_port_mappings/1, PortMappings);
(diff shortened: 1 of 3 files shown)


urbanserj:
@GoelDeepak request for review

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -586,30 +600,33 @@ merge_task_ports(PortMappings, PortResources, DiscoveryPorts) ->
...
+    [Port|Acc];
+merge_task_port(#{protocol := _P} = PortA, [PortB|Ports], Acc) ->
+    case match_task_port(PortA, PortB) of
+        true ->
+            PortC = merge_ports(PortB, PortA),
+            merge_task_port(PortC, [], Acc ++ Ports);
+        false ->
+            merge_task_port(PortA, Ports, [PortB|Acc])
+    end;
+merge_task_port(Port, Ports, Acc) ->
```
GoelDeepak: any reason why we are not ignoring the ports without `protocol`?

urbanserj: That's a good question, we can ignore them. Let me update the PR.

PR #69, opened 2018-08-20
[1.11] VIPs DNS is not removed even when application is deleted

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-3982

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_lashup_vip_listener.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_lashup_vip_listener.erl
@@ -137,9 +137,7 @@ handle_value(VIPs0, State0) ->
 process_vips(VIPs0, State) ->
     VIPs1 = lists:map(fun rewrite_keys/1, VIPs0),
     CategorizedVIPs = categories_vips(VIPs1),
-    RebindVIPs = [rebind_names(Family, VIPs, State)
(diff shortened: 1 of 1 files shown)


GoelDeepak:
@urbanserj The changes LGTM me. Could you please add a unit test? Also, would you know why jenkins is failing?

urbanserj:
@GoelDeepak I've added a test case for this bug. Jenkins build is failing on windows because 1.11 branch doesn't support it.

apps/dcos_l4lb/src/dcos_l4lb_lashup_vip_listener.erl
```diff
@@ -427,7 +425,10 @@ test_ip6(IP6, _, _) ->
     list_to_tuple(NextIP6).
 
 process_vips_tcp_test() ->
-    process_vips(tcp).
+    State = process_vips(tcp),
```
GoelDeepak: would it make sense to also assertNotEqual([], ets:tab2list(name_to_ip)) here?

urbanserj: There is an assert for that in `process_vip/1`.

https://github.com/dcos/dcos-net/pull/69/files#diff-599c9ad6aeffb3325a8c2108bcc67ed8R459

PR #75, opened 2018-09-06
[1.11] Add abort mode

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS-38600


urbanserj:
@GoelDeepak it's `disabled` by default:

PR #79, opened 2018-09-25
 Store tasks as LWW lashup values

urbanserj:
At the moment, DNS and VIPs records are stored as OR-Sets. This CRDT
data structure is based on vector clocks. Vector Clocks require lots
of memory and disk resources, the merge operation is very expensive.

Each dcos-net agent manages its own subset of mesos tasks with agent
ip as a key. All conflicts are resolved using last-write-wins (LWW)
strategy. This strategy is simple and doesn't require much cpu,
memory, or disk resources. Store the original data about all tasks
instead of storing processed data such as dns records and vip labels.
In the next releases, all submodules generate all records from this
lashup kv map.

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
@@ -95,6 +95,7 @@ handle_poll(Prev, true) ->
             Prev;
         {ok, Tasks} ->
             ok = handle_poll_state(Tasks),
+            ok = push_tasks(Tasks),
             Tasks
     catch error:bad_agent_id ->
         lager:warning("Mesos agent is not ready"),
@@ -256,6 +257,14 @@ push_ops(_Key, []) ->
     ok;
 push_ops(Key, Ops) ->
     {ok, _} = lashup_kv:request_op(Key, {update, Ops}),
+    ok.
+
+-spec(push_tasks(#{task_id() => task()}) -> ok).
+push_tasks(Tasks) ->
+    Ts = erlang:system_time(millisecond),
+    LKey = {dcos_net_dist:nodeip(), riak_dt_lwwreg},
+    Ops = [{update, LKey, {assign, Tasks, Ts}}],
+    {ok, _} = lashup_kv:request_op([tasks], {update, Ops}),
     ok.
 
 -spec(log_ops([riak_dt_map:map_field_update()]) -> ok).
--- a/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
@@ -17,7 +17,8 @@
     key/0, lkey/0, backend/0]).
 
 -record(state, {
-    timer_ref :: reference()
+    timer_ref :: reference(),
+    prev = #{} :: #{task_id() => task()}
 }).
 
 -type task() :: dcos_net_mesos_listener:task().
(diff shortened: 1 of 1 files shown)


GoelDeepak:
@urbanserj  Please don't merge it yet

GoelDeepak:
While I appreciate the thought but I feel it requires a deeper discussion to evaluate the pros and cons of this approach. I agree that the `OR-set` and vector clock are expensive and it might be possible to use LWW instead. However, pushing the entire tasks object to lashup might not be desirable on multiple fronts:
1. It is increasing both RAM and disk space on each agent. We need to estimate the worst case consumption for these resources
2. There would be a node to node transfer of this data. In the context of Hybrid cloud, where nodes might be in different data centers, we need to estimate what it would take to transfer 
3. Is this solution scalable? We need to estimate how the system resources would grow as we grow the number of nodes in the cluster, the number of tasks on each node, and the size of each task
4. what is the magnitude of the error under this design? Say there is some parser error. Currently, only the `dcos-net` that has a particular task experiences the parser error. Under this new design all `dcos-net` would experience that error.
5. With this approach, we are also making our implementation very tightly coupled with the task object. Going forward tasks may not be the only source of DNS or L4LB information. We have seen multiple requests for custom DNS. 
6. How about garbage collection? Especially, in the scenarios when an agent has retired. I know our current story is not too great either

It might be possible that we could still use LWW but for the processed data instead of keeping raw tasks list. Remember that we are also spending CPU cycles on each agent doing the same work

I understand the motivation to push it in DC/OS 1.12 so that we could handle the upgrade in next release. However, I feel we should first think deeply about the above questions (and more) before going in this direction. Unfortunately, we are very late in DC/OS 1.12 cycle.

urbanserj:
@GoelDeepak I totally understand your last point. Indeed, we are very late in DC/OS 1.12 cycle. Let me address all your other concerns for posterity:

> 1. It is increasing both RAM and disk space on each agent. We need to estimate the worst case consumption for these resources

dcos-dns generates 3 dns records for each task, also it creates copies for all mesos-dns records (two for each task and n srv reconds for its discovery ports), dcos-l4lb creates small vip record. size of dns records are x times bigger than pretty compact `task()` that contains only valuable information.

> 2. There would be a node to node transfer of this data. In the context of Hybrid cloud, where nodes might be in different data centers, we need to estimate what it would take to transfer

dns records are much bigger, but only leader node creates them; new LWW map can generate more traffic if lashup doesn't have a buffer or doesn't throttle multicast requests. 

> 3. Is this solution scalable? We need to estimate how the system resources would grow as we grow the number of nodes in the cluster, the number of tasks on each node, and the size of each task

Indeed, see my previous comment. It'll be a very good performance test for lashup.

> 4. what is the magnitude of the error under this design? Say there is some parser error. Currently, only the dcos-net that has a particular task experiences the parser error. Under this new design all dcos-net would experience that error.

New method saves the same data that l4lb poller receives from mesos agent. It's not something new, basically, we'll just save these data in another (raw) format. All components process mesos master/agent state using the same parser. The parser skips a bad task and logs an error message if there is an issue. In that case, we will lose only one task.  

> 5. With this approach, we are also making our implementation very tightly coupled with the task object. Going forward tasks may not be the only source of DNS or L4LB information. We have seen multiple requests for custom DNS.

It's okay. You can generate custom DNS records from another source and add them to `task()`-based DNS records. You would use exactly the same appoach even with current architecture.

> 6. How about garbage collection? Especially, in the scenarios when an agent has retired. I know our current story is not too great either

LWW map has agent ip as a key, it's much much easier to add GC for such data structures than for OR-Set with vector clocks. We can use LWW timestamp or agent list from mesos master as a source of trust for GC.

>  Remember that we are also spending CPU cycles on each agent doing the same work

I'm pretty sure it's not a problem for DNS records because it's very cheap to generate them from `task()` and it could have a better performance than coping huge blobs. For VIPs, it requires performance testing.

PR #81, opened 2018-10-02
Skip restablish vtep

mainred:
Reconfiging mac address of an interface will reset its static ip neighbours.
Thus, static ip neighbours of vtep interface will be removed,
if dcos_overlay is restarted, which results in traffic break.

Fix this by skipping configuring an existing vtep interface.
Also we delete and create the vtep interface if its configuration is   not up to date.

Close-Bug: DCOS_OSS-3855


mainred:
@urbanserj @GoelDeepak  PTAL

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -172,9 +199,44 @@ maybe_create_vtep(#{<<"backend">> := Backend}, #state{netlink = Pid}) ->
           ok = try_enable_ipv6(VTEPNameStr),
           {ParsedVTEPIP6, PrefixLen6} = parse_subnet(VTEPIP6),
           {ok, _} = dcos_overlay_netlink:ipaddr_replace(Pid, inet6, ParsedVTEPIP6, PrefixLen6, VTEPNameStr),
-          lager:notice("Overlay VTEP was configured, ~s => ~p", [VTEPName, Info#{ip => VTEPIP6}])
+          lager:notice("Overlay VTEP address was configured, ~s => ~p", [VTEPName, Info#{ipv6 => VTEPIP6}])
+    end.
+
+check_vtep_link(Pid, VXLan) ->
+    #{<<"vtep_name">> := VTEPName} = VXLan,
+    VTEPNameStr = binary_to_list(VTEPName),
```
urbanserj: don't convert `VTEPName` to string, use binary whenever it's possible.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -172,9 +199,44 @@ maybe_create_vtep(#{<<"backend">> := Backend}, #state{netlink = Pid}) ->
...
+    case dcos_overlay_netlink:iplink_show(Pid, VTEPNameStr) of
+        {ok, [#rtnetlink{type=newlink, msg=Msg}]} ->
+            {unspec, arphrd_ether, _, _, _, LinkInfo} = Msg,
+            Match = match_vtep_link(VXLan, LinkInfo),
+            {true, Match};
+        {error, ErrorCode, ResponseMsg} ->
+            lager:info("Failed to find ~p for error_code: ~p, msg: ~p", [VTEPNameStr, ErrorCode, ResponseMsg]),
```
urbanserj: Should it be `error` message? Could you please add context to the message?

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -172,9 +199,44 @@ maybe_create_vtep(#{<<"backend">> := Backend}, #state{netlink = Pid}) ->
...
+
+check_vtep_link(Pid, VXLan) ->
+    #{<<"vtep_name">> := VTEPName} = VXLan,
+    VTEPNameStr = binary_to_list(VTEPName),
+    case dcos_overlay_netlink:iplink_show(Pid, VTEPNameStr) of
+        {ok, [#rtnetlink{type=newlink, msg=Msg}]} ->
+            {unspec, arphrd_ether, _, _, _, LinkInfo} = Msg,
+            Match = match_vtep_link(VXLan, LinkInfo),
+            {true, Match};
+        {error, ErrorCode, ResponseMsg} ->
```
urbanserj: What does that function return if there is no such ip link?

GoelDeepak: Thats "No such device"

urbanserj: Let's much only match `{ok, [#rtnetlink{}]}` and `{error, 3992977407, []}` here. All other errors should explicitly (and automatically) crash the process. We don't really need logging in "no such device" case, it's initial state of the system.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -145,24 +146,50 @@ config_overlay(Overlay, State) ->
...
+            lager:notice("Attributes of ~p are not update-to-date, and vtep will be recreated.", [VTEPNameStr]),
+            dcos_overlay_netlink:iplink_delete(Pid, VTEPNameStr),
+            create_vtep_link(VXLan, State)
+    end,
+
+    create_vtep_addr(VXLan, State).
+
+maybe_create_vtep(#{<<"backend">> := Backend}, State) ->
+    #{<<"vxlan">> := VXLan} = Backend,
+    maybe_create_vtep_link(VXLan, State).
```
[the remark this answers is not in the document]
mainred: I agree with you on this point.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -145,24 +146,50 @@ config_overlay(Overlay, State) ->
...
     Info = #{vni => VNI, mac => VTEPMAC, attr => VTEPAttr},
-    lager:notice("Overlay VTEP was configured, ~s => ~p", [VTEPName, Info#{ip => VTEPIP}]),
+    lager:notice("Overlay VTEP link was configured, ~s => ~p", [VTEPName, Info]).
+
+maybe_create_vtep_link(VXLan, State=#state{netlink = Pid}) ->
+    #{<<"vtep_name">> := VTEPName} = VXLan,
+    VTEPNameStr = binary_to_list(VTEPName),
+    case check_vtep_link(Pid, VXLan) of
+        false ->
+            lager:notice("~p will be created.", [VTEPNameStr]),
```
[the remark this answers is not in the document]
mainred: I will use info to log for the check result , and notice for changes made

apps/dcos_overlay/src/dcos_overlay_poller.erl
mainred: This error_code is not standard one, but keep constant with varied interface name as I tested. It's ok to just match `not found`.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -172,9 +235,10 @@ maybe_create_vtep(#{<<"backend">> := Backend}, #state{netlink = Pid}) ->
           ok = try_enable_ipv6(VTEPNameStr),
           {ParsedVTEPIP6, PrefixLen6} = parse_subnet(VTEPIP6),
           {ok, _} = dcos_overlay_netlink:ipaddr_replace(Pid, inet6, ParsedVTEPIP6, PrefixLen6, VTEPNameStr),
-          lager:notice("Overlay VTEP was configured, ~s => ~p", [VTEPName, Info#{ip => VTEPIP6}])
+          lager:notice("Overlay VTEP address was configured, ~s => ~p", [VTEPName, Info#{ipv6 => VTEPIP6}])
```
urbanserj: `Info#{ip => VTEPIP6}` was a correct value. We don't use ipv4 address for ipv6 vtep addr, that's why `ip` must be overridden.

urbanserj: Also `Info` now contains only one `ip` value, so it would be better to simplify logging in `create_vtep_addr/2`.

```
lager:notice("Overlay VTEP address was configured, ~s => ~p", [VTEPName, VTEPIP]),
…
lager:notice("Overlay VTEP address was configured, ~s => ~p", [VTEPName, VTEPIP6])
```

mainred: It seems IPv4 is a must and ipv4 is optional.
It first created an ipv4 address and create a coexisting ipv6 address optionally? So I should log i configured a ipv4 address, and then i configured an ipv6 address.

urbanserj: yup, that's how it is right now, but your patch is removing [this](https://github.com/dcos/dcos-net/pull/81/files#diff-8c19cd0256d5151b9727cdde5bbb7de8L165) line (see it in split mode). it should be one log message for ipv4 and one log message for ipv6.

mainred:
@urbanserj  done

GoelDeepak:
Thanks, @mainred and @urbanserj for working on this and I liked the conversation that you two had. Also, +1 on logging

PR #82, opened 2018-10-02
Minor logging improvements

urbanserj:
Before:

```
diff([{key, [x, y]}], [{key, [y, x]}]) ->
  {[], [], [{key, [], []}]}.
```

After:
```
diff([{key, [x, y]}], [{key, [y, x]}]) ->
  {[], [], []}.
```


urbanserj:
@GoelDeepak I've added a test case. Could you please take a look? Thank you!

PR #85, opened 2018-10-09
Allow agent hostnames to be real DNS names

addisonj:
The hostname value of a host can be a real DNS name if you use
`--hostname` or `--hostname-lookup` when starting master and agents.

This was assuming that the hostname was an IP address and you would get
an unhandled einval exception when calling the inet ipv4 parse.

This changes this to properly resolve a hostname and handle the new
error cases, such as nxdomain and einval


apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -278,10 +278,13 @@ handle_agent(Obj) ->
...
-        {ok, IP} = inet:parse_ipv4strict_address(IPStr),
+        %% The hostname can be a DNS hostname, so we need to try and resolve it
+        %% but we handle an nxdomain or malformed hostname below
+        {ok, IP} = inet:getaddr(IPStr, inet),
         #{id => Id, ip => IP}
-    catch error:{badkey, _} ->
-        #{id => Id, ip => undefined}
+    catch error:{badkey, _} -> #{id => Id, ip => undefined};
+          error:{nxdomain, _} -> #{id => Id, ip => undefined};
+          error:{einval, _} -> #{id => Id, ip => undefined}
```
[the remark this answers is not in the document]
addisonj: ah okay, i was wondering about that, will fix

addisonj:
Cherry picked, which seems to take care of both those issues, thanks for the info as well! Help me slowly build up some erlang basics :smile:

apps/dcos_net/test/dns-hostname.json
```diff
@@ -0,0 +1 @@
..."source":"test.f12cc459-cb38-11e8-a264-2e431fa5cff2","discovery":{"visibility":"FRAMEWORK","name":"test","ports":{}}}}],"completed_executors":[]},"get_frameworks":{"frameworks":[{"framework_info":{"user":"root","name":"marathon","id":{"value":"d0a07cf9-7424-426b-b138-2be9d07b8e62-0001"},"failover_timeout":604800,"checkpoint":true,"role":"slave_public","hostname":"10.60.23.91","principal":"dcos_marathon","webui_url":"http://10.60.23.91:8080","capabilities":[{"type":"TASK_KILLING_STATE"},{"type":"GPU_RESOURCES"},{"type":"PARTITION_AWARE"},{"type":"REGION_AWARE"}]}}]},"get_agents":{"agents":[{"agent_info":{"hostname":"localhost","port":5051,"resources":[{"name":"ports","type":"RANGES","ranges":{"range":[{"begin":1025,"end":2180},{"begin":2182,"end":3887},{"begin":3889,"end":5049},{"begin":5052,"end":8079},{"begin":8082,"end":8180},{"begin":8182,"end":32000}]}},{"name":"disk","type":"SCALAR","scalar":{"value":13397.0}},{"name":"cpus","type":"SCALAR","scalar":{"value":4.0}},{"name":"mem","type":"SCALAR","scalar":{"value":14555.0}}],"id":{"value":"d0a07cf9-7424-426b-b138-2be9d07b8e62-S4"},"domain":{"fault_domain":{"region":{"name":"aws/us-east-1"},"zone":{"name":"aws/us-east-1d"}}}}}]}}}
```
ichernetsky: It would be nice to have the content of this file to be `jq .`ed.

apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -24,20 +19,62 @@ family(IP) when size(IP) == 8 ->
...
-resolve_mesos(DNSName) ->
-    Opts = [{nameservers, dcos_dns_config:mesos_resolvers()}],
-    case inet_res:resolve(DNSName, in, a, Opts, 1000) of
-        {ok, DnsMsg} ->
-            case inet_dns:msg(DnsMsg, anlist) of
-                [DnsRecord] ->
-                    {ok, inet_dns:rr(DnsRecord, data)};
-                _DnsRecords ->
-                    {error, not_found}
+imp_resolve(<<"localhost">>, inet, _Timeout) ->
```
ichernetsky: What about having the `inet6` case for `localhost`?

apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -24,20 +19,62 @@ family(IP) when size(IP) == 8 ->
...
+        case Family of
+            inet -> ?DNS_TYPE_A;
+            inet6 -> ?DNS_TYPE_AAAA
+        end,
+    Query = #dns_query{name = DNSName, type = Type},
+    Message = #dns_message{rd = true, qc = 1, questions = [Query]},
+    Request = dns:encode_message(Message),
+    case dcos_dns_handler:resolve(udp, Request, Timeout) of
+        {ok, Response} ->
+            try dns:decode_message(Response) of #dns_message{answers = RRs} ->
```
ichernetsky: There should be a newline after `of`.

apps/dcos_dns/src/dcos_dns.erl
urbanserj: There is no ipv6 address for `localhost` in dcos-dns:

https://github.com/dcos/dcos-net/blob/master/apps/dcos_dns/src/dcos_dns_sup.erl#L98-L110

apps/dcos_net/test/dns-hostname.json
urbanserj: We had troubles with reviewing very long json test data and switch to one liners in DC/OS 1.12.

apps/dcos_dns/src/dcos_dns.erl
urbanserj: IIRC, we have been using that code style to avoid linter errors because our config has very strict rule for that particular case. https://github.com/dcos/dcos-net/blob/6b41175/rebar.config#L193

PR #92, opened 2018-10-12
dcos_dns: make ZK servers configurable

ichernetsky:
JIRA issues: DCOS_OSS-4311

### DIFF
--- a/apps/dcos_dns/src/dcos_dns_key_mgr.erl
+++ b/apps/dcos_dns/src/dcos_dns_key_mgr.erl
@@ -6,16 +6,20 @@
 -endif.
 
 %% API
--export([start_link/0,
-         keys/0]).
+-export([
+         start_link/0,
+    keys/0
+]).
 
 %% gen_server callbacks
--export([init/1,
-         handle_call/3,
-         handle_cast/2,
-         handle_info/2,
-         terminate/2,
-         code_change/3]).
+-export([
+    init/1,
+    handle_call/3,
+    handle_cast/2,
+    handle_info/2,
+    terminate/2,
+    code_change/3
+]).
 
 -define(SERVER, ?MODULE).
 -define(LASHUP_CHECK_INTERVAL, 5000).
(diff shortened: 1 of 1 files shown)


apps/dcos_dns/src/dcos_dns_key_mgr.erl
```diff
@@ -199,14 +187,16 @@ push_data_to_lashup(#{public := Pk, secret := Sk}) ->
     true.
 
 get_zookeepers() ->
+    ZooKeeperServers = application:get_env(dcos_dns, zookeeper_servers,
+                                           ?DEFAULT_ZOOKEEPER_SERVERS),
```
urbanserj: nit: we'd prefer to avoid "standard" emacs indentation for erlang code. It would be just very painful to manage it and force that code style if there is someone who doesn't use such fancy plugins. We usually use 4 spaces everywhere with some exceptions, e.g. multiline list comprehensions.

apps/dcos_dns/src/dcos_dns_key_mgr.erl
```diff
@@ -1,41 +1,29 @@
...
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
 %% API
--export([
-    start_link/0,
-    keys/0
-]).
+-export([start_link/0,
```
GoelDeepak: Could we keep the previous format? This is something we try to follow

apps/dcos_dns/src/dcos_dns_key_mgr.erl
```diff
@@ -1,41 +1,29 @@
...
-    handle_call/3,
-    handle_cast/2,
-    handle_info/2,
-    terminate/2,
-    code_change/3]).
+         handle_call/3,
+         handle_cast/2,
+         handle_info/2,
+         terminate/2,
+         code_change/3]).
```
GoelDeepak: could we move ]) to next line. Again thats the style we are trying to follow

apps/dcos_dns/src/dcos_dns_key_mgr.erl
GoelDeepak: Styling: I would suggest to either keep all the arguments in one line (if possible) or move each argument in a separate line

apps/dcos_dns/src/dcos_dns_key_mgr.erl
```diff
@@ -6,16 +6,20 @@
 -endif.
 
 %% API
--export([start_link/0,
-         keys/0]).
+-export([
+         start_link/0,
```
GoelDeepak: nit: spaces is a little weird here

PR #93, opened 2018-10-12
Provide instructions on how to setup a local dcos-net cluster

ichernetsky:
JIRA issues: DCOS_OSS-4309


docs/dev.md
```diff
@@ -0,0 +1,420 @@
...
+To check your dcos-net build on DC/OS you can use DC/OS E2E a.k.a.
+[dcos-docker CLI](http://dcos-e2e.readthedocs.io/en/latest/cli.html):
+
+```sh
+curl -O https://downloads.dcos.io/dcos/testing/master/dcos_generate_config.sh
+make dcos-docker-create DCOS_DOCKER_AGENTS=3
+make dcos-docker-dev DCOS_DOCKER_NODE=master_0
+make dcos-docker-shell DCOS_DOCKER_NODE=agent_1
+make dcos-docker-destroy
+```
```
Fabs: What do you think about making the sessions clearly separated? between the simple way to do it and the advanced.

```md
## Basic Development
- curl -O https://downloads.dcos.io/dcos/testing/master/dcos_generate_config.sh
- make dcos-docker-create DCOS_DOCKER_AGENTS=3


## Advanced Development

Alternatively....
```

docs/dev.md
```diff
@@ -0,0 +1,420 @@
...
+
+```sh
+curl -LO https://raw.githubusercontent.com/mesosphere/marathon/master/tools/mvm.sh
+chmod +x mvm.sh
+```
+
+And build Mesos using it:
+
+```sh
+mvm.sh 1.7.0
```
Fabs: Maybe it pays of to comment that compiling mesos have a lot of dependencies, and point to this docs, because it helps you quickly find the dependencies for your platform (http://mesos.apache.org/documentation/latest/building/).

docs/dev.md
```diff
@@ -0,0 +1,420 @@
...
+         --vmethod MASQ -m comment \
+         --comment Minuteman-IPVS-IPTables-masquerade-rule \
+         -j MASQUERADE
+ip link add minuteman type dummy
+ip link set minuteman up
+ip link add spartan type dummy
+ip link set spartan up
+ip addr add 198.51.100.1/32 dev spartan
+ip addr add 198.51.100.2/32 dev spartan
+ip addr add 198.51.100.3/32 dev spartan
```
Fabs: I had trouble with that on a mac. I believe that this is due to trying to modprobe on a docker container on docker on mac. Somehow if you use dcos-e2e gets around that networking on a different way. Is it worth mentioning here how that can be done, or at least just mentioned that if you are on a mac with docker, you are out of luck?

docs/dev.md
```diff
@@ -0,0 +1,420 @@
+# Development
+
+You can build, check, and test dcos-net in a development image using
```
urbanserj: Could you please move `make docker*` and `make dcos-docker*` to `readme.md` and keep only advanced instruction here? It's pretty easy to start contributing to dcos-net with `make` commands, the steps are really simple and you should know almost nothing about dc/os or erlang infrastructure. So it would be good to advertise them loudly. Advanced instruction is for rare and specific cases.

docs/dev.md
urbanserj: @Fabs These commands should work fine if you run them in docker with `--privileged`.

Fabs: Hum, I will research a little more. I tried with `--privileged` before, but although I got some extra capabilities, `modprobe`, `rmmod` and etc was not available. Maybe I am missing something here. I will look more carefully. Thanks.

docs/dev.md
Fabs: Yes, having the advanced threw me off the tracks a bit, I could have simply used the `dcos-docker*`

docs/build.md
```diff
@@ -0,0 +1,406 @@
...
+NODE_2_IP=192.168.0.2
+```
+
+Firstly, the dependencies should be installed. Most of them can be installed
+using the package manager of your Linux distribution. In case Erlang 21.x is not
+available through it, please build one using [kerl](https://github.com/kerl/kerl):
+
+```sh
+KERL_CONFIGURE_OPTIONS=
+KERL_CONFIGURE_OPTIONS+="--enable-dirty-schedulers "
```
urbanserj: nit: dirty shedulers and kernel poll are enabled by default in OTP 21, IIRC it's not even possible to disable them in that version. also we don't use hipe, so it could be disabled.

ichernetsky: Are you sure? I have just checked its `configure` scripts, and it seems that both can be disabled and and dirty schedulers are disabled by default. I will remove the HiPe flag since it is unnecessary.

urbanserj: http://erlang.org/download/otp_src_21.0.readme

I didn't check that, but according the release notices:

1. It's not possible disable dirty schedulers when building erlang. You can disable them in runtime with `+SDio` 0 and `+SDcpu 0:0`, but io stops working.

```
  OTP-14613    Application(s): erts

               It is not longer possible to disable dirty schedulers
               when building erlang.
```

2. you can disable kernel-polling only in build, now I'm not sure if it's enabled by default.

```
  OTP-14346    Application(s): erts

               The I/O polling functionality of erts has been
               re-written to better make use of the OSs polling
               mechanisms. This change means that erts will now always
               prefer to use a kernel-polling mechanism if possible.
               Also all of the I/O polling has been moved to dedicated
               threads instead of being placed in the scheduler loops.

               As a result of this, the erl options +K and +secio have
               been removed. It is still possible to disable
               kernel-poll, but it has to be done at compile time
               through the configure option --disable-kernel-poll.

               The new erl options +IOt and +IOp can be used to change
               how many IO poll threads and poll sets that erts should
               use. See their respective documentation for more
               details.
```

All of these options work fine with OTP 21, so we can keep all of them in the instructions.

docs/build.md
```diff
@@ -0,0 +1,405 @@
+# Building from the ground up
```
mainred: We have many steps and sub steps to build a dcos-net cluster, do you mind adding subtitles to make it more organized, if it can?

## preparation
## compile erlang
## compile DC/OS Mesos module
## start up the components
Exhibitor, Mesos-master and Mesos-DNS are deployed on master node, and Mesos-agent are deployed on agent node, and dcos-net runs on both nodes.
### start Exhibitor (master node)
### start Mesos-master (master node)
### start Mesos-DNS (master node)
### start Mesos-agent (agent node)
### start dcos-net
1.  ...
2.  ...

PR #97, opened 2018-10-20
re-write overlay test

GoelDeepak:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-4286

### DIFF
--- a/apps/dcos_overlay/test/dcos_overlay_SUITE.erl
+++ b/apps/dcos_overlay/test/dcos_overlay_SUITE.erl
@@ -1,14 +1,4 @@
-%%%-------------------------------------------------------------------
-%%% @author dgoel
-%%% @copyright (C) 2016, <COMPANY>
-%%% @doc
-%%%
-%%% @end
-%%% Created : 31. Oct 2016 11:07 AM
-%%%-------------------------------------------------------------------
 -module(dcos_overlay_SUITE).
--author("dgoel").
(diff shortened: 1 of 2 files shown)


apps/dcos_overlay/test/dcos_overlay_SUITE.erl
```diff
@@ -179,126 +77,53 @@ meck_httpc_request(get, {_, Headers}, _, [{sync, false}]) ->
...
-  Result = ordsets:subtract(ordsets:from_list(Expected), ordsets:from_list(Actual)),
-  io:format("Result ~p\n", [Result]),
-  Result.
-
-expected_data(Master) ->
-  NumNodes = ?AGENT_COUNT + ?MASTER_COUNT,
-  MasterNode = parse_node(list_to_binary(atom_to_list(Master))),
-  [ get_node_data(N) || N <- lists:seq(1, NumNodes), N =/= MasterNode].
+    Nodes = ?config(nodes, Config),
+    {_, []} = rpc:multicall(Nodes, meck, new,
```
urbanserj: @GoelDeepak could you please try to move all `meck:new/2` and `meck:expect/2` calls to new `meck_setup/1` function and call that function on all `Nodes` using `erlang:apply/2`?

```
rpc:multicall(Nodes, erlang, apply, [fun meck_setup/1, []])
```

PR #98, opened 2018-10-26
dcos-l4lb routes traffic until the first failed health check

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-1954
Apache Mesos RR: https://reviews.apache.org/r/69110/


apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -430,19 +432,23 @@ id2bin({FrameworkId, TaskId}) ->
...
         _TaskState ->
-            true
+            preparing
     end.
 
+-spec(handle_task_healthy(jiffy:object(), task()) -> undefined | boolean()).
+handle_task_healthy(TaskObj, _Task) ->
+    IsHealthy =
+        case maps:is_key(<<"health_check">>, TaskObj) of
+            false -> undefined;
```
GoelDeepak: wouldn't "undefined" break the lists:filter on is_healthy later? We should make it "true" to fall back to the previous behavior.

apps/dcos_net/test/dcos_net_mesos_listener_tests.erl
```diff
@@ -111,7 +112,28 @@ tcp_udp_test() ->
                 #{name => <<"foobar">>, protocol => udp,
                   port => 6417}
             ],
-            state => {running, true}
+            state => running,
+            healthy => true
+        }
+    }, dcos_net_mesos_listener:from_state(Data)).
+
+unhealthy_test() ->
```
GoelDeepak: Would it make sense to add few more "negative" test cases? For instance:
1. state is preparing and healthy is true  -- it shouldn't happen but our code shouldn't break either
2. state = '_' and health_check undefined -- as it use to happen earlier. 
3. state = terminal and healthy is true

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: `undefined` wouldn't break any code. In that case `mput/2` doesn't update task info and it doesn't have `healthy` field.

apps/dcos_net/test/dcos_net_mesos_listener_tests.erl
urbanserj: 1. We'd prefer to use a real mesos state, but not to generate random data

2. There are a bunch of tests for case 2.

3. it is a regular case.

PR #101, opened 2018-11-14
Improvements and bug fixes in dns forwarder

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS-45161

### DIFF
--- a/apps/dcos_dns/src/dcos_dns_handler.erl
+++ b/apps/dcos_dns/src/dcos_dns_handler.erl
@@ -166,11 +166,23 @@ init_worker(tcp, Pid, Upstreams, Request) ->
 -spec(udp_worker(pid(), upstream(), binary()) -> ok).
 udp_worker(Pid, Upstream = {IP, Port}, Request) ->
     StartTime = erlang:system_time(millisecond),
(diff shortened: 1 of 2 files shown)


apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -165,7 +165,7 @@ init_worker(tcp, Pid, Upstreams, Request) ->
 
 -spec(udp_worker(pid(), upstream(), binary()) -> ok).
 udp_worker(Pid, Upstream = {IP, Port}, Request) ->
-    StartTime = erlang:system_time(millisecond),
+    StartTime = erlang:monotonic_time(millisecond),
     MonRef = erlang:monitor(process, Pid),
     Opts = [{reuseaddr, true}, {active, once}, binary],
     {ok, Socket} = gen_udp:open(0, Opts),
```
GoelDeepak: Should will wrap UDP in a `case` as well?

urbanserj: Done. I've also wrapped all `send/2` calls.

PR #102, opened 2018-11-20
Changes the backend for a VIP in case of docker containerizer

GoelDeepak:
In case of docker containerizer if the backend to a VIP is the
actual container ip then the application running in the container
cannot access its own VIP. The only way to make it work is to go
through port mapping. This patch fixes the port mapping such that
the backend is the host IP:port intead of container ip:port

jira: DCOS-45115

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
@@ -124,10 +124,12 @@ is_healthy(_Task) ->
          Container :: {inet:ip_address(), inet:port_number()}).
 collect_port_mappings(Tasks) ->
     maps:fold(fun (_TaskId, Task, Acc) ->
+        Runtime = maps:get(runtime, Task),
         [TaskIP | _TaskIPs] = maps:get(task_ip, Task),
         PMs = [{{Protocol, Host}, {TaskIP, Port}}
               || #{host_port := Host, protocol := Protocol,
-                   port := Port} <- maps:get(ports, Task, [])],
+                   port := Port} <- maps:get(ports, Task, []),
+                 Runtime =/= docker],
         PMs ++ Acc
     end, [], Tasks).
 
--- a/apps/dcos_l4lb/test/dcos_l4lb_mesos_poller_SUITE.erl
+++ b/apps/dcos_l4lb/test/dcos_l4lb_mesos_poller_SUITE.erl
@@ -1,19 +1,26 @@
 -module(dcos_l4lb_mesos_poller_SUITE).
 
+-include_lib("eunit/include/eunit.hrl").
 -include_lib("common_test/include/ct.hrl").
 -include("dcos_l4lb.hrl").
 
 -export([
     all/0,
     init_per_suite/1, end_per_suite/1,
-    init_per_testcase/2, end_per_testcase/2,
-    test_lashup/1
+    init_per_testcase/2, end_per_testcase/2
+]).
+
+-export([
+    test_lashup/1,
+    test_mesos_portmapping/1
 ]).
 
 
 %% root tests
-all() ->
-    [test_lashup].
+all() -> [
+    test_lashup,
+    test_mesos_portmapping
+].
 
 init_per_suite(Config) ->
     Config.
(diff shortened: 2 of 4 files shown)


apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -21,6 +21,7 @@
 -type task() :: #{
     name => binary(),
     framework => binary() | {id, binary()},
+    container => binary() | undefined,
```
urbanserj: Let's rename it to `runtime`, make it required, and use `docker` and `mesos` atoms as values?

```suggestion
    runtime => docker | mesos,
```

apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
```diff
@@ -125,9 +125,15 @@ is_healthy(_Task) ->
...
     maps:fold(fun (_TaskId, Task, Acc) ->
         [TaskIP | _TaskIPs] = maps:get(task_ip, Task),
-        PMs = [{{Protocol, Host}, {TaskIP, Port}}
-              || #{host_port := Host, protocol := Protocol,
-                   port := Port} <- maps:get(ports, Task, [])],
+        AgentIP = maps:get(agent_ip, Task),
+        PMs = [{{Protocol, Host},
+                case maps:get(container, Task) of
+                    <<"DOCKER">> ->
+                         {AgentIP, Host};
```
urbanserj: It's not a port mapping. To fix the issue, we should use port mappings only for mesos runtime, i.e.

```
Runtime = maps:get(runtime, Task),
[ {{Protocol, Host}, {TaskIP, Port}}
|| #{host_port := Host, protocol := Protocol,
     port := Port} <- maps:get(ports, Task, []),
   Runtime =/= docker ],
```

apps/dcos_net/src/dcos_net_mesos_listener.erl
GoelDeepak: This doesn't work as certain events such as TASK_UPDATE doesn't have this information.

apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
GoelDeepak: Good suggestion!

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: It's expected, `TASK_UPDATE` returns task_status, not task info. In that case we just shouldn't update `runtime` field.

GoelDeepak:
@urbanserj  Could you please take a look now?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -21,6 +21,7 @@
 -type task() :: #{
     name => binary(),
     framework => binary() | {id, binary()},
+    runtime => binary(),
```
urbanserj: Could you please use `mesos | docker` atoms as a type for `runtime` field? Atoms consume 1 word, binaries consume 3-6 words + data size [1]. Code with atoms, in general, looks much cleaner. Also when you convert binaries to atoms, it's an additional validation of data from mesos.

If you think that it could be other container types, please consider adding the third atom (e.g. `unknown`).

[1] http://erlang.org/doc/efficiency_guide/advanced.html#memory

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -454,6 +456,10 @@ handle_task_name(TaskObj, _Task) ->
     Name = mget(<<"name">>, TaskObj, undefined),
     mget([<<"discovery">>, <<"name">>], TaskObj, Name).
 
+-spec(handle_task_runtime(jiffy:object(), task()) -> binary() | undefined).
+handle_task_runtime(TaskObj, _Task) ->
+    mget([<<"container">>, <<"type">>], TaskObj, undefined).
```
urbanserj: If the function returns undefined, the field is optional and this code [1] can throw `badkey` exception. I'd like to make `runtime` field required.

[1] https://github.com/dcos/dcos-net/pull/102/files#diff-b85318f2b911cad717b851556fb2153dR127

apps/dcos_l4lb/test/dcos_l4lb_mesos_poller_SUITE.erl
```diff
@@ -23,23 +30,14 @@ end_per_suite(Config) ->
...
-                task_ip => [{10, 0, 0, 243}],
-                ports => [
-                    #{name => <<"http">>, protocol => tcp,
-                      port => 12049, vip => [<<"merp:5000">>]}
-                ],
-                state => running
-            }
-        }}
-    end),
+    meck:expect(dcos_net_mesos_listener, poll, meck_expect_fn()),
```
urbanserj: My suggestion:

```
meck:expect(dcos_net_mesos_listener, poll, fun meck_mesos_poll/1),
…
meck_mesos_poll() ->
    {ok, #{…}}.
```

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj:
```
Default = maps:get(runtime, Task, unknown),
try mget([<<"container">>, <<"type">>], TaskObj) of
    <<"MESOS">> ->
        mesos;
    <<"DOCKER">> ->
        docker;
    Type ->
        lager:warning("", [TaskId, Type]),
        unknown
catch error:{badkey, _} ->
    Default
end
```

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -21,12 +21,14 @@
...
     name => binary(),
     framework => binary() | {id, binary()},
+    runtime => runtime(),
     agent_ip => inet:ip4_address() | {id, binary()},
     task_ip => [inet:ip_address()],
     state => task_state(),
     healthy => boolean(),
     ports => [task_port()]
 }.
+-type runtime() :: docker | mesos | unknown.
```
urbanserj: Could you please add `runtime/0` to the `export_type` list?

PR #104, opened 2018-11-21
Expose Public IPs

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-4514
DC/OS PR: https://github.com/dcos/dcos/pull/3796

### DIFF
--- a/apps/dcos_net/src/dcos_net_node.erl
+++ b/apps/dcos_net/src/dcos_net_node.erl
@@ -1,0 +1,260 @@
+-module(dcos_net_node).
+-behaviour(gen_server).
+
+%% API
+-export([
+    start_link/0,
+    get_metadata/0,
+    get_metadata/1
+]).
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2,
+    handle_info/2, terminate/2, code_change/3]).
+
+-export_type([metadata/0]).
+
+-type metadata() :: #{
+    public_ips => [inet:ip_address()],
+    hostname => binary(),
+    updated => non_neg_integer()
+}.
+
+-record(state, {
+    timer_ref :: reference(),
+    lashup_ref :: reference(),
+    refresh = #{} :: map()
+}).
+-type state() :: #state{}.
+
+-include_lib("stdlib/include/ms_transform.hrl").
+-define(LASHUP_KEY, [dcos_net, nodes]).
+
+-spec(start_link() ->
+    {ok, Pid :: pid()} | ignore | {error, Reason :: term()}).
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
+
+-spec(get_metadata() -> #{inet:ip4_address() => metadata()}).
+get_metadata() ->
+    List = lashup_kv:value(?LASHUP_KEY),
+    dump(List).
+
+-spec(get_metadata(proplists:proplist()) -> #{inet:ip4_address() => metadata()}).
+get_metadata([]) ->
+    get_metadata();
+get_metadata([force_refresh]) ->
+    % It's not possible to unsubscribe from lashup_kv_events_helper's updates.
+    % force_dump/0 function starts a new process that does all the work, sends
+    % the result back and terminates itself.
+    % TODO: Add unsubscribe to the events helper and get rid of spawn_link.
+    Ref = make_ref(),
+    PPid = self(),
+    proc_lib:spawn_link(fun() -> force_refresh(PPid, Ref) end),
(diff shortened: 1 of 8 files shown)


apps/dcos_rest/src/dcos_rest_nodes_handler.erl
```diff
@@ -0,0 +1,56 @@
...
+    jiffy:json_object()).
+get_node(PrivateIP, Metadata) ->
+    PublicIPs = maps:get(public_ips, Metadata),
+    Hostname = maps:get(hostname, Metadata),
+    Updated = maps:get(updated, Metadata),
+
+    #{private_ip => ntoa(PrivateIP),
+      public_ips => [ntoa(IP) || IP <- PublicIPs],
+      hostname => Hostname,
+      updated => iso8601(Updated)}.
```
nLight: ❤️

apps/dcos_net/src/dcos_net_node.erl
```diff
@@ -0,0 +1,258 @@
...
+dump() ->
+    List = lashup_kv:value(?LASHUP_KEY),
+    dump(List).
+
+-spec(force_dump() -> #{inet:ip4_address() => metadata()}).
+force_dump() ->
+    % It's not possible to unsubscribe from lashup_kv_events_helper's updates.
+    % force_dump/0 function starts a new process that does all the work, sends
+    % the result back and terminates itself.
+    % TODO: Add ubsibscribe to the events helper and get rid of spawn_link.
```
Fabs: typo: ubsibscribe/unsubscribe

apps/dcos_net/src/dcos_net_node.erl
```diff
@@ -0,0 +1,258 @@
...
+    {ok, State}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+-spec(start_update_timer() -> reference()).
+start_update_timer() ->
+    Timeout = application:get_env(
+        dcos_net, update_node_info, timer:minutes(30)),
```
Fabs: have you considered the tradeofs of making this configurable form day 0? I wonder if we would want to be able to tweak and test this time without touching code.

Fabs: Maybe same way we did with `detect_ip_public`

apps/dcos_rest/src/dcos_rest_nodes_handler.erl
Fabs: image

apps/dcos_net/src/dcos_net_node.erl
urbanserj: `detect_ip_public` runs locally and doesn't depend on the size of a cluster, so we can use a reasonable constant as its timeout. On big clusters force refresh can take more time than on small ones. We don't have the goal to have 100% code coverage, force refresh and this timer run exactly the same code.

apps/dcos_net/src/dcos_net_node.erl
```diff
@@ -0,0 +1,258 @@
+-module(dcos_net_node).
+-behaviour(gen_server).
+
+%% API
+-export([
+    start_link/0,
+    dump/0,
+    force_dump/0
```
GoelDeepak: nit: Similarly instead of `force_dump/0` to something say `force_refresh/0`? I know there is another `force_refresh/0`, however, that is an internal function which we can name something else as compare to this one which is externally visible

apps/dcos_net/test/dcos_net_node_SUITE.erl
```diff
@@ -0,0 +1,53 @@
...
+    Config.
+
+end_per_suite(Config) ->
+    Config.
+
+init_per_testcase(_, Config) ->
+    {ok, _} = application:ensure_all_started(mnesia),
+    {ok, _} = application:ensure_all_started(lashup),
+    {ok, Pid} = dcos_net_node:start_link(),
+    timer:sleep(100),
```
GoelDeepak: do we need this sleep even when we have Wait()?

apps/dcos_net/test/dcos_net_node_SUITE.erl
```diff
@@ -0,0 +1,53 @@
...
+    (fun Wait() ->
```
GoelDeepak: I would make it an explicit function call for readability

apps/dcos_net/src/dcos_net_node.erl
urbanserj: I've added `get_metadata/1` function with `force_refresh` option.

PR #110, opened 2019-01-10
[1.11] ssl_tls_dist_proxy hang if distribution protocol over TLS connections time out

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS-46915
JIRA: https://jira.mesosphere.com/browse/SOAK-157

### DIFF
--- a/apps/dcos_net/src/dcos_net_killer.erl
+++ b/apps/dcos_net/src/dcos_net_killer.erl
@@ -43,6 +43,7 @@ handle_info({timeout, Ref, kill}, #state{ref=Ref, reductions=Prev}=State) ->
                 _Other -> Acc
             end
         end, [], Rs),
+    true = maybe_kill_dist_proxy(),
     lists:foreach(fun maybe_kill/1, StuckProcesses),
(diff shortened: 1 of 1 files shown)


apps/dcos_net/src/dcos_net_killer.erl
```diff
@@ -70,6 +71,41 @@ reductions() ->
...
+    case erlang:process_info(Pid, messages) of
+        {messages, Messages} ->
+            maybe_kill_dist_proxy(Pid, Messages);
+        undefined ->
+            true
+    end.
+
+-spec(maybe_kill_dist_proxy(pid(), [term()]) -> true | no_return()).
+maybe_kill_dist_proxy(Pid, Messages) ->
+    MaxMsgs = application:get_env(dcos_net, ssl_tls_dist_proxy_max_msgs, 64),
```
GoelDeepak: What is the rationale behind 64 messages?

urbanserj: On soak cluster `ssl_tls_dist_proxy` receives a connection request every 10 seconds or so when it hits this issue. So in 10-15 minutes the killer will stop the whole vm and systemd restart it.

PR #111, opened 2019-01-11
Add `thisnode.thisdcos.directory` dns zone

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-4666
JIRA: https://jira.mesosphere.com/browse/DCOS-46999

### DIFF
--- a/apps/dcos_dns/src/dcos_dns.erl
+++ b/apps/dcos_dns/src/dcos_dns.erl
@@ -1,6 +1,8 @@
 -module(dcos_dns).
 
+-include("dcos_dns.hrl").
 -include_lib("dns/include/dns_records.hrl").
+-include_lib("erldns/include/erldns.hrl").
 
 %% API
 -export([
@@ -9,6 +11,18 @@
     resolve/2,
     resolve/3,
     get_leader_addr/0
+]).
+
+%% DNS Zone functions
+-export([
+    ns_record/1,
+    soa_record/1,
+    dns_record/2,
+    dns_records/2,
+    srv_record/2,
+    cname_record/2,
+    push_zone/2,
+    push_prepared_zone/2
 ]).
 
 -spec family(inet:ip_address()) -> inet | inet6.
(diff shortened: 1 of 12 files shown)


apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -78,3 +92,109 @@ imp_resolve(DNSName, Family, Timeout) ->
...
+                {error, Error} ->
+                    lager:error(
+                        "Failed to push DNS Zone \"~s\": ~p",
+                        [ZoneName, Error]),
+                    {error, Error}
+            end
+    catch error:Error ->
+        lager:error(
+            "Failed to push DNS Zone \"~s\": ~p",
+            [ZoneName, Error]),
```
GoelDeepak: Do we have a line number in our error logs? If not, then probably we might want to have a slightly different error message just to differentiate from the error message at lineno. 185

GoelDeepak:
What is the purpose of `distance` change?

apps/dcos_dns/src/dcos_dns.erl
urbanserj: There is only one possible error with `erldns_zone_cache:get_zone_with_records/1`. It's `error(badarg)`, when erldns cache is not initialized and that's only possible in tests.

GoelDeepak: 👍

urbanserj:
@GoelDeepak I've moved `distance` to new branch [1]. it's a small change, but it's definitely out of the scope for this PR. We can discuss it separately later.

https://github.com/dcos/dcos-net/compare/urbanserj/nodes-distance

GoelDeepak:
> @GoelDeepak I've moved `distance` to new branch [1]. it's a small change, but it's definitely out of the scope for this PR. We can discuss it separately later.
> 
> https://github.com/dcos/dcos-net/compare/urbanserj/nodes-distance

Make sense. Thank you for splitting into two different PRs. It will give us more time to brainstorm the `distance` approach

PR #120, opened 2019-02-25
Improving initialization time

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-4828

Initialization of prometheus library with default `instrumenters` configuration can take up to 2 minutes if dcos-net is started by `rebar3 shell` or using release build in dev mode. This patch fixes the issue.

```
1> timer:tc(fun() -> prometheus_instrumenter:enabled_instrumenters() end).
{78677677,[prometheus_test_instrumenter]}

vs.

1> timer:tc(fun() -> prometheus_instrumenter:enabled_instrumenters() end).
{25,[]}
```


config/sys.config
```diff
@@ -104,6 +104,7 @@
             prometheus_vm_statistics_collector,
             prometheus_vm_system_info_collector
         ]},
+        {instrumenters, []},
```
Fabs: I am curious to know what does this exactly do? Does that remove a bunch of instruments that come by default?

urbanserj: If the list of `instrumenters` is not configured, prometheus library loads all modules and checks which of them have `behaviour` attribute set to `prometheus_instrumenter`. The code loading takes a lot of time. At the moment there is only one such module, `prometheus_test_instrumenter`.

https://github.com/deadtrickster/prometheus.erl/blob/master/src/prometheus_instrumenter.erl#L43
https://github.com/deadtrickster/prometheus.erl/blob/master/src/prometheus_misc.erl

PR #121, opened 2019-03-07
Add metrics for dns workers

Fabs:
JIRA: https://jira.mesosphere.com/browse/DCOS-49018

### DIFF
--- a/apps/dcos_dns/src/dcos_dns_handler.erl
+++ b/apps/dcos_dns/src/dcos_dns_handler.erl
@@ -177,90 +177,106 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
 -spec(start_worker(protocol(), upstream(), binary()) -> pid()).
 start_worker(Protocol, Upstream, Request) ->
     Pid = self(),
+    Begin = erlang:monotonic_time(),
+    BeginMilliseconds = erlang:monotonic_time(millisecond),
     proc_lib:spawn(fun () ->
-        init_worker(Protocol, Pid, Upstream, Request)
+        UpstreamAddress = upstream_to_binary(Upstream),
+        case init_worker(Protocol, Pid, Upstream, Request) of
+            {ok, Response} ->
+                Pid ! {reply, self(), Response},
+                report_latency([?MODULE, Upstream, latency], BeginMilliseconds),
+                prometheus_histogram:observe(
+                    dns, worker_requests_duration_seconds,
+                    [Protocol, UpstreamAddress],
+                    erlang:monotonic_time() - Begin);
+            {error, Reason} ->
+                lager:warning(
+                    "DNS worker [~p] ~p failed with ~p",
+                    [Protocol, UpstreamAddress, Reason]),
+                prometheus_counter:inc(
+                    dns, worker_failures_total,
+                    [Protocol, UpstreamAddress, Reason], 1)
+        end
     end).
 
--spec(init_worker(protocol(), pid(), upstream(), binary()) -> ok).
-init_worker(udp, Pid, Upstreams, Request) ->
-    udp_worker(Pid, Upstreams, Request);
-init_worker(tcp, Pid, Upstreams, Request) ->
-    tcp_worker(Pid, Upstreams, Request).
+-spec(init_worker(protocol(), pid(), upstream(), binary()) ->
+    {ok, binary()} | {error, atom()}).
+init_worker(udp, Pid, Upstream, Request) ->
+    udp_worker(Pid, Upstream, Request);
+init_worker(tcp, Pid, Upstream, Request) ->
+    tcp_worker(Pid, Upstream, Request).
 
--spec(udp_worker(pid(), upstream(), binary()) -> ok).
-udp_worker(Pid, Upstream = {IP, Port}, Request) ->
-    StartTime = erlang:monotonic_time(millisecond),
+-spec(udp_worker(pid(), upstream(), binary()) ->
+    {ok, binary()} | {error, atom()}).
+udp_worker(Pid, {IP, Port}, Request) ->
     Opts = [{reuseaddr, true}, {active, once}, binary],
     case gen_udp:open(0, Opts) of
         {ok, Socket} ->
             case gen_udp:send(Socket, IP, Port, Request) of
                 ok ->
-                    udp_worker(StartTime, Pid, Socket, Upstream);
-                {error, Error} ->
-                    lager:warning("DNS worker ~p failed with ~p", [Upstream, Error])
+                    udp_worker_receive(Pid, {IP, Port}, Socket);
+                {error, Reason} ->
+                    {error, Reason}
             end;
-        {error, Error} ->
-            lager:warning("DNS worker ~p failed with ~p", [Upstream, Error])
+        {error, Reason} ->
+            {error, Reason}
     end.
 
--spec(udp_worker(integer(), pid(), gen_udp:socket(), upstream()) -> ok).
-udp_worker(StartTime, Pid, Socket, Upstream = {IP, Port}) ->
+-spec(udp_worker_receive(pid(), upstream(), gen_udp:socket()) ->
+    {ok, binary()} | {error, atom()}).
+udp_worker_receive(Pid, {IP, Port}, Socket) ->
     MonRef = erlang:monitor(process, Pid),
     try
         receive
-            {'DOWN', MonRef, process, _Pid, normal} ->
-                ok;
-            {'DOWN', MonRef, process, _Pid, Reason} ->
-                exit(Reason);
+            {'DOWN', MonRef, process, _Pid, _Reason} ->
+                {error, down};
             {udp, Socket, IP, Port, Response} ->
-                Pid ! {reply, self(), Response},
-                report_latency([?MODULE, Upstream, latency], StartTime);
+                {ok, Response};
             {timeout, Pid} ->
-                lager:warning("DNS worker ~p timed out", [Upstream])
+                {error, timeout}
         after ?TIMEOUT ->
-            ok
+            {error, timeout}
         end
     after
         gen_udp:close(Socket)
     end.
 
--spec(tcp_worker(pid(), upstream(), binary()) -> ok).
-tcp_worker(Pid, Upstream = {IP, Port}, Request) ->
+-spec(tcp_worker(pid(), upstream(), binary()) ->
+   {ok, binary()} | {error, atom()}).
+tcp_worker(Pid, {IP, Port}, Request) ->
     StartTime = erlang:monotonic_time(millisecond),
     Opts = [{active, once}, binary, {packet, 2}, {send_timeout, ?TIMEOUT}],
     case gen_tcp:connect(IP, Port, Opts, ?TIMEOUT) of
         {ok, Socket} ->
             case gen_tcp:send(Socket, Request) of
                 ok ->
-                    tcp_worker(StartTime, Pid, Socket, Upstream);
-                {error, Error} ->
-                    lager:warning("DNS worker [tcp] ~p failed with ~p", [Upstream, Error])
+                    tcp_worker_receive(StartTime, Pid, Socket);
+                {error, Reason} ->
+                    {error, Reason}
             end;
-        {error, Error} ->
-            lager:warning("DNS worker [tcp] ~p failed with ~p", [Upstream, Error])
+        {error, Reason} ->
+            {error, Reason}
     end.
 
--spec(tcp_worker(integer(), pid(), gen_tcp:socket(), upstream()) -> ok).
-tcp_worker(StartTime, Pid, Socket, Upstream) ->
+-spec(tcp_worker_receive(integer(), pid(), gen_tcp:socket()) ->
+   {ok, binary()} | {error, atom()}).
+tcp_worker_receive(StartTime, Pid, Socket) ->
     MonRef = erlang:monitor(process, Pid),
     try
         Timeout = ?TIMEOUT - (erlang:monotonic_time(millisecond) - StartTime),
         receive
-            {'DOWN', MonRef, process, _Pid, normal} ->
-                ok;
-            {'DOWN', MonRef, process, _Pid, Reason} ->
-                exit(Reason);
+            {'DOWN', MonRef, process, _Pid, _Reason} ->
+                {error, down};
             {tcp, Socket, Response} ->
-                Pid ! {reply, self(), Response},
-                report_latency([?MODULE, Upstream, latency], StartTime);
+                {ok, Response};
             {tcp_closed, Socket} ->
-                lager:warning("DNS worker [tcp] ~p failed with ~p", [Upstream, closed]);
-            {tcp_error, Socket, Reason} ->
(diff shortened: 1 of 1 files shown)


apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,89 +177,107 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
-udp_worker(StartTime, Pid, Socket, Upstream = {IP, Port}) ->
+-spec(udp_worker(integer(), pid(), gen_udp:socket(), {upstream(), binary()}) -> ok).
+udp_worker(StartTime, Pid, Socket, {Upstream = {IP, Port}, Address}) ->
     MonRef = erlang:monitor(process, Pid),
     try
         receive
             {'DOWN', MonRef, process, _Pid, normal} ->
                 ok;
             {'DOWN', MonRef, process, _Pid, Reason} ->
+                emit_worker_failure(Address, udp, Reason),
```
Fabs: The `Reason` from gen_tcp/gen_udp is of `inet:posix()` but the Reason here is from exit and can be `term()`. That would mess up with `prometheus`. I am thinking that we might need here to check when it is not an atom, and just get the first atom of the tuple... but not really sure how to solve that. @urbanserj, Ideas?

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,89 +177,107 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
+-spec(tcp_worker(integer(), pid(), gen_tcp:socket(), {upstream(), binary()}) -> ok).
+tcp_worker(StartTime, Pid, Socket, {Upstream, Address}) ->
     MonRef = erlang:monitor(process, Pid),
     try
         Timeout = ?TIMEOUT - (erlang:monotonic_time(millisecond) - StartTime),
         receive
             {'DOWN', MonRef, process, _Pid, normal} ->
                 ok;
             {'DOWN', MonRef, process, _Pid, Reason} ->
+                emit_worker_failure(Address, tcp, Reason),
```
Fabs: ditto

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -379,4 +403,22 @@ init_metrics() ->
...
+        {name, worker_failures_total},
+        {labels, [upstream_address, protocol, reason]},
+        {help, "Total number of worker failures processing DNS requests."}]),
+    prometheus_histogram:new([
+        {registry, dns},
+        {name, worker_requests_duration_seconds},
+        {labels, [upstream_address, protocol]},
+        {duration_unit, seconds},
+        {buckets, [0.001, 0.005, 0.010, 0.050, 0.100, 0.500, 1.000, 5.000]},
+        {help, "The time spent a worker spent processing DNS requests."}]).
```
Fabs: naming is not good.

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -315,6 +333,12 @@ upstream_failures(Upstream) ->
             Failures
     end.
 
+-spec(upstream_to_binary(upstream()) -> binary()).
```
Fabs: The other examples are similar, but they actualy do `a name concat with port`, and not ip. So they are simpler. I decided to live this as is. Will though write at least a test.

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,89 +177,107 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
+        {error, Reason} ->
+            lager:warning("DNS worker ~p failed with ~p", [Address, Reason]),
+            emit_worker_failure(Address, udp, Reason)
     end.
 
--spec(udp_worker(integer(), pid(), gen_udp:socket(), upstream()) -> ok).
-udp_worker(StartTime, Pid, Socket, Upstream = {IP, Port}) ->
+-spec(udp_worker(integer(), pid(), gen_udp:socket(), {upstream(), binary()}) -> ok).
+udp_worker(StartTime, Pid, Socket, {Upstream = {IP, Port}, Address}) ->
     MonRef = erlang:monitor(process, Pid),
```
Fabs: Also maybe something is still confusing for me. This Pid here is `self()` before we call spawn for the worker. So I am assuming the worker monitors its parent.

- Why is it not necessary to demonitor before the worker goes away?
- Should we really emit this failure here when the parent has an issue?

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,89 +177,107 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
-            lager:warning("DNS worker ~p failed with ~p", [Upstream, Error])
+        {error, Reason} ->
+            lager:warning("DNS worker ~p failed with ~p", [Address, Reason]),
+            emit_worker_failure(Address, udp, Reason)
     end.
 
--spec(udp_worker(integer(), pid(), gen_udp:socket(), upstream()) -> ok).
-udp_worker(StartTime, Pid, Socket, Upstream = {IP, Port}) ->
+-spec(udp_worker(integer(), pid(), gen_udp:socket(), {upstream(), binary()}) -> ok).
+udp_worker(StartTime, Pid, Socket, {Upstream = {IP, Port}, Address}) ->
```
urbanserj: We can simplify the logic here and move all logging and metrics to the parent function. All we need are to return `{ok, Response :: binary()} | {error, Error :: atom()}.`, i.e.

```
-spec(udp_worker(pid(), gen_udp:socket()) -> {ok, binary()} | {error, atom()}).
udp_worker(Pid, Socket) ->
    MonRef = erlang:monitor(process, Pid),
    receive
        {'DOWN', MonRef, process, Pid, _Reason} ->
            {error, down};
        {udp, Socket, _IP, _Port, Response} ->
            {ok, Response};
        {timeout, Pid} ->
            {error, timeout}
    after ?TIMEOUT ->
        {error, timeout}
    end.
```

```
case udp_worker(Pid, Socket) of
    {ok, Response} ->
        Pid ! {reply, self(), Response};
    {error, Error} ->
        lager:warning(…),
        prometheus_counter:inc(…);
end,
gen_udp:close(Socket);
```

apps/dcos_dns/src/dcos_dns_handler.erl
urbanserj: > Why is it not necessary to demonitor before the worker goes away?

Because either worker or parent process will be terminated in a jiffy. In worse case, worker will just have one `'DOWN'` unprocessed message in its mailbox. We can `demonitor` parent pid with `[flush]` in `after` section, but basically it doesn't change anything.

> Should we really emit this failure here when the parent has an issue?

Hmm, it's a good question. It's not an upstream dns server issue, so probably no.

apps/dcos_dns/src/dcos_dns_handler.erl
urbanserj: Let's ignore it, basically `normal` and `Reason` is the same expected case. It's not an upstream error.

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -379,4 +403,22 @@ init_metrics() ->
...
+        {help, "Total number of worker failures processing DNS requests."}]),
+    prometheus_histogram:new([
+        {registry, dns},
+        {name, worker_requests_duration_seconds},
+        {labels, [upstream_address, protocol]},
+        {duration_unit, seconds},
+        {buckets, [0.001, 0.005, 0.010, 0.050, 0.100, 0.500, 1.000, 5.000]},
+        {help, "The time a worker spent processing DNS requests."}]).
+
+-spec(emit_worker_failure(binary(), udp | tcp, atom()) -> ok).
```
urbanserj: Could you please try to move the logic from receive functions to parent functions and maybe in that case we don't have to introduce this one?

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,89 +177,107 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
 -spec(start_worker(protocol(), upstream(), binary()) -> pid()).
 start_worker(Protocol, Upstream, Request) ->
     Pid = self(),
+    Begin = erlang:monotonic_time(),
     proc_lib:spawn(fun () ->
-        init_worker(Protocol, Pid, Upstream, Request)
+        Address = upstream_to_binary(Upstream),
+        init_worker(Protocol, Pid, {Upstream, Address}, Request),
```
urbanserj: We ever can move all metrics and logging here and make all other functions much cleaner, right?

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,90 +177,100 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
 -spec(start_worker(protocol(), upstream(), binary()) -> pid()).
 start_worker(Protocol, Upstream, Request) ->
     Pid = self(),
+    Begin = erlang:monotonic_time(),
+    BeginMilliseconds = erlang:monotonic_time(millisecond),
```
Fabs: the `BeginMilliseconds` will be removed on the next PR when we clean up the `report_latency`

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,90 +177,106 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
     proc_lib:spawn(fun () ->
```
urbanserj: This lambda function has started being too big.

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,90 +177,106 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
+                    [Protocol, UpstreamAddress, Reason], 1)
+        end
     end).
 
--spec(init_worker(protocol(), pid(), upstream(), binary()) -> ok).
-init_worker(udp, Pid, Upstreams, Request) ->
-    udp_worker(Pid, Upstreams, Request);
-init_worker(tcp, Pid, Upstreams, Request) ->
-    tcp_worker(Pid, Upstreams, Request).
+-spec(init_worker(protocol(), pid(), upstream(), binary()) ->
```
urbanserj: There were 3 functions that tail called each other. At the moment, `init_worker/4` does much more than just worker initialization. Could you please rename it to `worker_resolve(…)` or something like this?

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,90 +177,106 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
+                {ok, Response};
             {timeout, Pid} ->
-                lager:warning("DNS worker ~p timed out", [Upstream])
+                {error, timeout}
         after ?TIMEOUT ->
-            ok
+            {error, timeout}
         end
     after
         gen_udp:close(Socket)
```
urbanserj: With new semantics, I'd recommend to move `gen_udp:close/1` call to `udp_worker/3` function.

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,90 +177,106 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
+                {error, tcp_error};
             {timeout, Pid} ->
-                lager:warning("DNS worker [tcp] ~p timed out", [Upstream])
+                {error, timeout}
         after Timeout ->
-            ok
+            {error, timeout}
         end
     after
         gen_tcp:close(Socket)
```
urbanserj: Same as `gen_udp:close/1`.

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,90 +177,106 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
+            {ok, Response} ->
+                Pid ! {reply, self(), Response},
+                report_latency([?MODULE, Upstream, latency], BeginMilliseconds),
+                prometheus_histogram:observe(
+                    dns, worker_requests_duration_seconds,
+                    [Protocol, UpstreamAddress],
+                    erlang:monotonic_time() - Begin);
+            {error, Reason} ->
+                lager:warning(
+                    "DNS worker [~p] ~p failed with ~p",
```
urbanserj:
```suggestion
                    "DNS worker [~p] ~s failed with ~p",
```

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,90 +177,106 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
-            lager:warning("DNS worker [tcp] ~p failed with ~p", [Upstream, Error])
+        {error, Reason} ->
+            {error, Reason}
     end.
 
--spec(tcp_worker(integer(), pid(), gen_tcp:socket(), upstream()) -> ok).
-tcp_worker(StartTime, Pid, Socket, Upstream) ->
+-spec(tcp_worker_receive(integer(), pid(), gen_tcp:socket()) ->
+   {ok, binary()} | {error, atom()}).
+tcp_worker_receive(StartTime, Pid, Socket) ->
```
urbanserj: It looks like `udp_worker_receive/3` and `tcp_worker_receive/3` have almost the same logic, let merge them into one function?

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -177,90 +177,106 @@ resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
...
+    Begin = erlang:monotonic_time(),
+    BeginMilliseconds = erlang:monotonic_time(millisecond),
     proc_lib:spawn(fun () ->
-        init_worker(Protocol, Pid, Upstream, Request)
+        UpstreamAddress = upstream_to_binary(Upstream),
+        case init_worker(Protocol, Pid, Upstream, Request) of
+            {ok, Response} ->
+                Pid ! {reply, self(), Response},
+                report_latency([?MODULE, Upstream, latency], BeginMilliseconds),
+                prometheus_histogram:observe(
```
urbanserj: Could you please collect this metric for `error` case as well?

apps/dcos_dns/src/dcos_dns_handler.erl
Fabs: Do you think the fix is to extract it as a named function?

urbanserj: Yup, it's just a bit easier to read. Also all anonymous functions (since OTP 22, anonymous functions that capture the environment) are created in runtime.

Fabs:
@urbanserj. I read your code and welcome the improvements. Thank you. 

I have merged it, tested a bit this morning and I think it is ready for you to look again :).

PR #123, opened 2019-03-11
Remove the old DNS metrics

Fabs:
JIRA: https://jira.mesosphere.com/browse/DCOS-49018

It is getting cooler - https://gist.github.com/Fabs/0c6a216fdaed07922ee1dddd431f6f8b
Screenshot from 2019-03-12 17-30-06


apps/dcos_dns/src/dcos_dns_router.erl
```diff
@@ -12,15 +12,23 @@
...
+        find_upstream(Labels)
+    end,
+    Answers = lists:map(UpstreamAnswers, Questions),
+
+    case lists:usort(Answers) of
+        [Answer] ->
+            Answer;
+        MixedAnswers ->
+            InternalResolver = {internal, <<".">>},
+            lager:notice("DNS question ~p yield multiple results ~p,
```
ichernetsky: nit: does `lager` do something special to remove extra space chars? The Erlang parser don't:

```
~ $ cat ./multi-line 
#!/usr/bin/env escript
%% -*- erlang -*-

main([]) ->
    io:format("a
        b~n").
~ $ ./multi-line 
a
        b
```

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -151,21 +146,18 @@ resolve_loop(Begin, [], Fun, FailMsg, Zone) ->
 resolve_loop(Begin, Workers, Fun, FailMsg, Zone) ->
     receive
         {reply, Pid, Response} ->
-            {value, {Upstream, Pid, MonRef}, Workers0} =
+            {value, {_Upstream, Pid, MonRef}, Workers0} =
```
ichernetsky: Would it make sense to use `Upstream` as a metrics tag, since the number of upstream servers is small hence it won't make the metrics cardinality much bigger?

apps/dcos_dns/src/dcos_dns_router.erl
Fabs: Hum, interesting. I saw this message on the screen and it looked fine for me. I probably overlooked it. let me check again

Fabs: Yeah this is broken:
```
01:09:53.720 [notice] A
          B~n
```

apps/dcos_dns/src/dcos_dns_handler.erl
Fabs: If I am not mistaken, the same query for a DNS address (e.g. google.com) could be forwarded to different upstream (eg, 8.8.8.8:53 and 8.8.4.4:53). How would we capture that as a label? 

We do send the `upstream_address` which is an `ip + port` of the Upstream for `worker` metrics (forwarding launches a worker per upstream).  There my guess is that cardinality should not be a problem, as I would expect the number of distinct upstream addresses to be in the order of `10`.

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -308,14 +293,22 @@ choose(N, List) ->
...
+    UpstreamFailuresFilterMap = upstream_failures_filtermap_fun(Upstream),
+    AllFailures = prometheus_counter:values(dns, worker_failures_total),
+    UpstreamFailures = lists:filtermap(UpstreamFailuresFilterMap, AllFailures),
+    lists:sum(UpstreamFailures).
+
+-spec(upstream_failures_filtermap_fun(upstream()) ->
+   fun(({nonempty_maybe_improper_list(), number()}) -> false | {true, number()})).
+upstream_failures_filtermap_fun(Upstream) ->
+    QueryAddress = upstream_to_binary(Upstream),
+    fun({[{"upstream_address", Address} | _], Failures}) ->
```
urbanserj: It's broken here.

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -308,14 +293,22 @@ choose(N, List) ->
     List2 = [X || {_, X} <- List1],
     lists:sublist(List2, N).
 
--spec(upstream_failures([upstream()]) -> non_neg_integer()).
+-spec(upstream_failures(upstream()) -> number()).
```
urbanserj: Prometheus library stores all metrics using much different data structure. I'd recommend to re-write `take_upstreams/2` function considering this knowledge, e.g.

```erlang
-spec(take_upstreams([upstream()]) -> [upstream()]).
take_upstreams(Upstreams) when length(Upstreams) < 3 ->
    Upstreams;
take_upstreams(Upstreams) ->
    Buckets = upstream_buckets(Upstreams),
    case lists:unzip(Buckets) of
        {_N, [[_, _ | _] = Bucket | _Buckets]} ->
            choose(2, Bucket);
        {_N, [[Upstream], Bucket | _Buckets]} ->
            [Upstream | choose(1, Bucket)]
    end.

-spec(upstream_failures() -> orddict:orddict(upstream(), pos_integer())).
upstream_failures() ->
    AllFailures = prometheus_counter:values(dns, worker_failures_total),
    Failures = [{Address, Count} || {Labels, Count} <- AllFailures,
                                    {"upstream_address", Address} <- Labels],
    lists:foldl(fun ({Address, Count}, Acc) ->
        orddict:update_counter(Address, Count, Acc)
    end, orddict:new(), Failures).

-spec(upstream_buckets([upstream()]) ->
    orddict:orddict(non_neg_integer(), [upstream()])).
upstream_buckets(Upstreams) ->
    Failures = upstream_failures(),
    lists:foldl(fun (Upstream, Acc) ->
        Address = upstream_to_binary(Upstream),
        Count =
            case orddict:find(Address, Failures) of
                {ok, N} -> N;
                error -> 0
            end,
        orddict:append_list(Count, [Upstream], Acc)
    end, orddict:new(), Upstreams).
```

apps/dcos_dns/src/dcos_dns_router.erl
```diff
@@ -12,15 +12,22 @@
 %% @doc Resolvers based on a set of "questions"
 -spec(upstreams_from_questions(dns:questions()) ->
     {[upstream()] | internal, binary()}).
-upstreams_from_questions([#dns_query{name=Name}]) ->
```
urbanserj: It's the most common case, please let's keep it as is.

```
upstreams_from_questions([#dns_query{name=Name}]) ->
    Labels = dcos_dns_app:parse_upstream_name(Name),
    find_upstream(Labels);
upstreams_from_questions(Questions) ->
    AllUpstreams = [upstreams_from_questions([Q]) || Q <- Questions],
    case lists:usort(AllUpstreams) of
        [Upstream] ->
```

apps/dcos_dns/src/dcos_dns_router.erl
```diff
@@ -12,15 +12,22 @@
...
-    Labels = dcos_dns_app:parse_upstream_name(Name),
-    find_upstream(Labels);
-upstreams_from_questions([Question|Others]) ->
-    %% There is more than one question. This is beyond our capabilities at the moment
-    dcos_dns_metrics:update([dcos_dns, ignored_questions], length(Others), ?COUNTER),
-    Result = upstreams_from_questions([Question]),
-    lager:debug("~p will be forwarded to ~p", [Others, Result]),
-    Result.
+upstreams_from_questions(Questions) ->
+    UpstreamAnswers = fun (#dns_query{name=Name}) ->
```
urbanserj: DNS Answer means something different, please avoid such naming.

apps/dcos_dns/src/dcos_dns_router.erl
```diff
@@ -12,15 +12,22 @@
...
+        find_upstream(Labels)
+    end,
+    Answers = lists:map(UpstreamAnswers, Questions),
+
+    case lists:usort(Answers) of
+        [Answer] ->
+            Answer;
+        MixedAnswers ->
+            InternalResolver = {internal, <<".">>},
+            lager:notice("DNS question ~p has mixed upstream ~p. forwarding ~p",
```
urbanserj: `Questions` is a list of huge erlang terms, it's not human readable.

urbanserj: `notice` is used when we are changing some state, please use `warning` or `error in such cases.

urbanserj: nit:

```
lager:foobar(
    FormatString,
    Args)
```

apps/dcos_dns/src/dcos_dns_router.erl
```diff
@@ -12,15 +12,22 @@
...
+                [Questions, MixedAnswers, InternalResolver]),
+            InternalResolver
```
urbanserj: If it returns `{internal, []}`, we should update `forwarder_failures_total` metric for all dns zones except internal ones.

urbanserj:
IIRC `dcos_dns` is the last application that uses `exometer` library. Could you please remove the dependency from `rebar.config` and `rebar.lock`?

apps/dcos_dns/src/dcos_dns_handler.erl
Fabs: That is interesting. I am pretty sure that worked before, wit the same custom resolvers we used on my firs PR. I think at the end of the last PR I changed the order of the labels, so upstream address is not the first anymore `{labels, [protocol, upstream_address, reason]}`. That would have matched and then this error would not have happened.

I agree though that it is more robust to write code that does not depend on the order of the labels, as I already made a change that broke it :).

apps/dcos_dns/src/dcos_dns_handler.erl
Fabs: This line is really awesome:
```
    Failures = [{Address, Count} || {Labels, Count} <- AllFailures,
                                    {"upstream_address", Address} <- Labels]
```

rebar.lock
```diff
@@ -44,7 +43,7 @@
   {git,"https://github.com/dcos/mesos_state.git",
        {ref,"c4f23ea76cbf6a4cfaa534c2c42d22b02260778c"}},
   0},
- {<<"parse_trans">>,{pkg,<<"parse_trans">>,<<"3.3.0">>},1},
+ {<<"parse_trans">>,{pkg,<<"parse_trans">>,<<"3.3.0">>},2},
```
Fabs: This was changed by the rebar tool.

apps/dcos_dns/src/dcos_dns_router.erl
urbanserj: @Fabs I mean, you shouldn't create `FormatString` variable here, it's only about formatting:

```
OK:
module:function(A, B, C)

OK:
module:function(
    A,
    B,
    C)

NOT OK:
module:function(A,
    B, C)

NOT OK:
module:function(A,
    B,
    C)
```

Fabs:
@urbanserj done all the comments. I tested it with the same resolvers you proposed. I have not found a way to test queries with multiple questions, so I manually tested by injecting this in the `upstream_from-questions` and firing some equal queries, and different queries as well. 

```
Questions = QuestionsB ++ [{dns_query, <<"ready.spartan">>,1,1}],
```

Fabs:
Hum, the CI fails with a style issue, fixing that....

apps/dcos_dns/src/dcos_dns_router.erl
```diff
@@ -15,12 +15,33 @@
...
-    dcos_dns_metrics:update([dcos_dns, ignored_questions], length(Others), ?COUNTER),
-    Result = upstreams_from_questions([Question]),
-    lager:debug("~p will be forwarded to ~p", [Others, Result]),
-    Result.
+upstreams_from_questions(Questions) ->
+    AllUpstreams = [upstreams_from_questions([Q]) || Q <- Questions],
+    case lists:usort(AllUpstreams) of
+        [Upstream] ->
+            Upstream;
+        MixedUpstreams ->
```
urbanserj: Let's make the message more readable. I think it's fine to print erlang terms at the end of the message (in this case erlang term can be truncated, but human readable message will be printed). Here are my suggestion:

```erlang
        _Upstreams ->
            lager:warning(
                "DNS queries with mixed-upstream questions are not supported, "
                "the query will be resolved through internal DNS server: ~p",
                [Questions]),
            lists:foreach(fun (Zone) ->
                prometheus_counter:inc(
                    dns, forwarder_failures_total,
                    [Zone], 1)
            end, [Z || {U, Z} <- AllUpstreams, U =/= internal]),
            {internal, <<".">>}
```

apps/dcos_dns/src/dcos_dns_handler.erl
urbanserj: Oh, I overlooked this one. It seems like nobody uses the first element of this tuple. If so, could you please remove it?

Fabs:
@ichernetsky I just implement changes asked by Sergey, and also answered your feedback. If you would like to check again it is ready :).

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -151,21 +146,18 @@ resolve_loop(Begin, [], Fun, FailMsg, Zone) ->
...
+            {value, {Pid, MonRef}, Workers0} =
+                lists:keytake(Pid, 1, Workers),
             _ = reply(Begin, Fun, Response, Zone),
             erlang:demonitor(MonRef, [flush]),
-            report_status([?MODULE, Upstream, successes]),
             resolve_loop(Begin, Workers0, skip, FailMsg, Zone);
         {'DOWN', MonRef, process, Pid, _Reason} ->
-            {value, {Upstream, Pid, MonRef}, Workers0} =
+            {value, {Pid, MonRef}, Workers0} =
                 lists:keytake(Pid, 2, Workers),
```
urbanserj:
```suggestion
                lists:keytake(Pid, 1, Workers),
```

Fabs: Oh. I did that fix, it crashes otherwise. :persevere:

Fabs: I will squash again on Github if you don't mind.

PR #127, opened 2019-03-19
Add metrics for DNS Zones

Fabs:
JIRA: https://jira.mesosphere.com/browse/DCOS-49352

### DIFF
--- a/apps/dcos_dns/src/dcos_dns.erl
+++ b/apps/dcos_dns/src/dcos_dns.erl
@@ -10,7 +10,8 @@
     resolve/1,
     resolve/2,
     resolve/3,
-    get_leader_addr/0
+    get_leader_addr/0,
+    init_metrics/0
 ]).
 
 %% DNS Zone functions
@@ -170,23 +171,12 @@ push_zone(ZoneName, Records) ->
 -spec(push_prepared_zone(dns:dname(), [dns:dns_rr()]) ->
     ok | {error, Reason :: term()}).
 push_prepared_zone(ZoneName, Records) ->
+    Begin = erlang:monotonic_time(),
     try erldns_zone_cache:get_zone_with_records(ZoneName) of
         {ok, #zone{records=Records}} ->
             ok;
         _Other ->
-            Hash = crypto:hash(sha, term_to_binary(Records)),
-            case erldns_zone_cache:put_zone({ZoneName, Hash, Records}) of
-                ok ->
-                    lager:notice(
-                        "DNS Zone ~s was updated (~p records, sha: ~s)",
-                        [ZoneName, length(Records), bin_to_hex(Hash)]),
-                    ok;
-                {error, Error} ->
-                    lager:error(
-                        "Failed to push DNS Zone \"~s\": ~p",
-                        [ZoneName, Error]),
-                    {error, Error}
-            end
+            put_zone(Begin, ZoneName, Records)
     catch error:Error ->
         lager:error(
             "Failed to push DNS Zone \"~s\": ~p",
@@ -194,7 +184,57 @@ push_prepared_zone(ZoneName, Records) ->
         {error, Error}
     end.
 
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
 -spec(bin_to_hex(binary()) -> binary()).
 bin_to_hex(Bin) ->
     Bin0 = << <<(integer_to_binary(N, 16))/binary>> || <<N:4>> <= Bin >>,
     cowboy_bstr:to_lower(Bin0).
+
+-spec(put_zone(integer(), binary(), [dns:dns_rr()]) -> ok).
+put_zone(Begin, ZoneName, Records) ->
+    Hash = crypto:hash(sha, term_to_binary(Records)),
+    case erldns_zone_cache:put_zone({ZoneName, Hash, Records}) of
+        ok ->
+            Duration = erlang:monotonic_time() - Begin,
+            ZoneName0 = <<ZoneName/binary, ".">>,
+            lager:notice(
+                "DNS Zone ~s was updated (~p records, sha: ~s, duration: ~pms)",
+                [ZoneName0, length(Records), bin_to_hex(Hash),
+                erlang:convert_time_unit(Duration, native, millisecond)]),
+            try
+                prometheus_summary:observe(
+                    dns, zone_push_duration_seconds,
+                    [ZoneName0], Duration),
+                prometheus_gauge:set(
+                    dns, zone_records,
+                    [ZoneName0], length(Records))
+            catch error:_Error ->
+                    ok
+            end;
+        {error, Error} ->
+            lager:error(
+                "Failed to push DNS Zone \"~s\": ~p",
+                [ZoneName, Error]),
+            {error, Error}
+    end.
+
+%%%===================================================================
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_gauge:new([
+        {registry, dns},
+        {name, zone_records},
+        {labels, [zone]},
+        {help, "The number of records in DNS Zone."}]),
+    prometheus_summary:new([
+        {registry, dns},
+        {name, zone_push_duration_seconds},
+        {labels, [zone]},
+        {duration_unit, seconds},
+        {help, "The time spent pushing DNS Zone to the zone cache."}]).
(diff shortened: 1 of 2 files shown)


apps/dcos_dns/test/dcos_dns_mesos_tests.erl
```diff
@@ -356,6 +356,9 @@ pod_tasks_setup() ->
     setup(pod_tasks_setup).
 
 setup(SetupFun) ->
+    prometheus:start(),
```
Fabs: To make the tests work I added this here, and also changed the metrics declaration from `new` to `declare` as the latter does not require the metrics not to exist before.

Do you see a better way to solve for the fact that the tests exercise the function that calls `Prometheus`?

Fabs: Another weird observation is that locally the tests pass but now on CI, tests for the other metrics fail.

Fabs:
I know about this issue @urbanserj, I should have mentioned that. I am testing by building dashboards on grafana. 
Screenshot from 2019-03-20 09-29-41

By those two graphs, one can see that there are zones where we can see the `size` but there is no `duration` metrics.

I opened this PR with the `help wanted` label because of my issue with the tests. I looked at your `lashup` PR, and at other places but I was unable to find any instance where we had to deal with Prometheus and the ETS tables in the test. Do you have any suggestions there? Do you think it is related to my issue of missing push duration?

Fabs:
I know instrumented all calls to the `push_prepared_zone`.

Screenshot from 2019-03-20 12-51-32

apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -189,12 +198,58 @@ push_prepared_zone(ZoneName, Records) ->
...
+        {help, "Number of DNS Zone records"}]),
+    prometheus_summary:new([
+        {registry, dns},
+        {name, zone_push_duration_seconds},
+        {labels, [zone]},
+        {duration_unit, seconds},
+        {help, "The time spent pushing DNS Zones."}]).
+
+-spec(safe_summary_observe(atom(), atom(), [any()], any()) -> ok).
+safe_summary_observe(Registry, Metric, Labels, Value) ->
```
Fabs: I tried to do it without `try` but unit tests still fail.

Fabs:
@urbanserj, @ichernetsky could you give another pass?

apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -189,12 +198,58 @@ push_prepared_zone(ZoneName, Records) ->
...
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
 -spec(bin_to_hex(binary()) -> binary()).
 bin_to_hex(Bin) ->
     Bin0 = << <<(integer_to_binary(N, 16))/binary>> || <<N:4>> <= Bin >>,
     cowboy_bstr:to_lower(Bin0).
+
+-spec(emit_metrics(binary(), integer(), [dns:dns_rr()]) -> ok).
```
urbanserj: As we discussed earlier, please don't create metric functions when it's possible. In this case, I'd recommend to add `put_zone/2` function that will push data to erldns cache and do all logging and metric stuff, i.e. something like this (I didn't test it):

```erlang
-spec(push_prepared_zone(dns:dname(), [dns:dns_rr()]) ->
    ok | {error, Reason :: term()}).
push_prepared_zone(ZoneName, Records) ->
    Begin = erlang:monotonic_time(),
    try
        case erldns_zone_cache:get_zone_with_records(ZoneName) of
            {ok, #zone{records=Records}} ->
                ok;
            _Other ->
                put_zone(Begin, ZoneName, Records)
        end
    catch error:Error ->
        lager:error("Failed to push DNS Zone ~s: ~p", [ZoneName, Error]),
        {error, Error}
    end.

%%% Internal functions

-spec(put_zone(Begin :: integer(), dns:dname(), [dns:dns_rr()]) -> ok).
put_zone(Begin, ZoneName, Records) ->
    Hash = crypto:hash(sha, term_to_binary(Records)),
    case erldns_zone_cache:put_zone({ZoneName, Hash, Records}) of
        ok ->
            Duration = erlang:monotonic_time() - Begin,
            ZoneName0 = <<ZoneName/binary, ".">>,
            lager:notice(
                "DNS Zone ~s was updated (~p records, sha: ~s, duration: ~pms)",
                [ZoneName0, length(Records), bin_to_hex(Hash),
                 erlang:convert_time_unit(Duration, native, millisecond)]),
            prometheus_gauge:set(
                dns, zone_records,
                [ZoneName0], length(Records)),
            prometheus_summary:observe(
                dns, zone_push_duration_seconds,
                [ZoneName0], Duration);
        {error, Error} ->
            error(Error)
    end.
```

apps/dcos_dns/src/dcos_dns.erl
urbanserj: You actually don't need these two functions, they are called only once.

apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -189,12 +198,58 @@ push_prepared_zone(ZoneName, Records) ->
...
+        {registry, dns},
+        {name, zone_records},
+        {labels, [zone]},
+        {help, "Number of DNS Zone records"}]),
+    prometheus_summary:new([
+        {registry, dns},
+        {name, zone_push_duration_seconds},
+        {labels, [zone]},
+        {duration_unit, seconds},
+        {help, "The time spent pushing DNS Zones."}]).
```
urbanserj:
```suggestion
        {help, "The time spent pushing DNS Zone to the zone cache."}]).
```

apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -189,12 +198,58 @@ push_prepared_zone(ZoneName, Records) ->
...
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_gauge:new([
+        {registry, dns},
+        {name, zone_records},
+        {labels, [zone]},
+        {help, "Number of DNS Zone records"}]),
```
urbanserj:
```suggestion
        {help, "The number of records in DNS Zone."}]),
```

apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -170,16 +171,24 @@ push_zone(ZoneName, Records) ->
...
             Hash = crypto:hash(sha, term_to_binary(Records)),
             case erldns_zone_cache:put_zone({ZoneName, Hash, Records}) of
                 ok ->
+                    Elapsed = erlang:monotonic_time() - Begin,
+                    ElapsedUs = erlang:convert_time_unit(
+                        Elapsed, native, microsecond),
                     lager:notice(
-                        "DNS Zone ~s was updated (~p records, sha: ~s)",
-                        [ZoneName, length(Records), bin_to_hex(Hash)]),
+                        "DNS Zone ~s. was updated "
```
urbanserj: If you want to log `ZoneName` with `.` at the end, please do it explicitly in function arguments, not in the format string.

apps/dcos_dns/src/dcos_dns.erl
```diff
@@ -170,31 +171,68 @@ push_zone(ZoneName, Records) ->
...
-                        "Failed to push DNS Zone \"~s\": ~p",
-                        [ZoneName, Error]),
-                    {error, Error}
-            end
+            put_zone(Begin, ZoneName, Records)
     catch error:Error ->
         lager:error(
-            "Failed to push DNS Zone \"~s\": ~p",
-            [ZoneName, Error]),
+            "Failed to push DNS Zone \"~s\": ~p", [ZoneName, Error]),
```
urbanserj: nit: for consistency, could you please use one code style? either of them is fine

https://github.com/dcos/dcos-net/pull/127/files#diff-19215963023a2b45a6e2cd51922592d6R217

Fabs: Yeah. I also just noticed that I broke the lines inconsistently on the Prometheus metrics call. 

Is it possible that we codify those on a linter? Coming from JS, I am really used to just have this being done automatically for me by a tool...

urbanserj: Unfortunately, [elvis](https://github.com/inaka/elvis_core/blob/master/src/elvis_style.erl) doesn't support such things.

Fabs:
I fixed (I guess) both style error @urbanserj

urbanserj:
@Fabs LGTM, could you please squash all the commits?

PR #130, opened 2019-03-20
Fix a race condition in L4LB

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-4939
DC/OS PR: https://github.com/dcos/dcos/pull/4830

There are three components in VIP configuration:

1. `l4lb.thisdcos.directory` DNS Zone
2. Local routing table
3. IP_VS configuration

It is currently configured exactly in this order. It causes a race condition when DNS name is resolvable, VIP is routable to the loopback interface, but ip_vs configuration is not yet updated. In this case, TCP/UDP client connects to localhost service that is already running on VIP port using VIP address.

Here is how it should work:

1. Remove old unused routes
2. Update IP_VS configuration
3. Add new routes
4. Wait for steps [1-3]
5. Push DNS Zone to erldns cache

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
@@ -54,7 +54,11 @@
     when Key :: dcos_l4lb_mesos_poller:key(),
          Backend :: dcos_l4lb_mesos_poller:backend()).
 push_vips(VIPs) ->
-    gen_server:cast(?MODULE, {vips, VIPs}).
+    try
+        gen_server:call(?MODULE, {vips, VIPs})
+    catch exit:{noproc, _MFA} ->
+        ok
+    end.
 
 -spec(push_netns(EventType, [netns()]) -> ok
     when EventType :: add_netns | remove_netns | reconcile_netns).
@@ -75,11 +79,11 @@ init([]) ->
     self() ! init,
     {ok, []}.
 
+handle_call({vips, VIPs}, _From, State) ->
+    {reply, ok, handle_vips(VIPs, State)};
 handle_call(_Request, _From, State) ->
     {noreply, State}.
 
-handle_cast({vips, VIPs}, State) ->
-    {noreply, handle_vips(VIPs, State)};
 handle_cast({netns, Event}, State) ->
     {noreply, handle_netns_event(Event, State)};
 handle_cast(_Request, State) ->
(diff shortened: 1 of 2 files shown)


urbanserj:
`test_networking.test_dcos_cni_l4lb` has started failing with this patch 🤔

urbanserj:
All integration tests passed!

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -182,45 +182,61 @@ start_reconcile_timer() ->
...
+        ok = add_routes(RouteMgr, RoutesToAdd, Namespace),
+        ok = log_routes_diff(LogPrefix, {RoutesToAdd, []})
     end, Namespaces),
     State#state{prev_ipvs=VIPs1, prev_routes=Routes}.
 
--spec(handle_vips_imp([{key(), [backend()]}], state()) -> state()).
-handle_vips_imp(VIPs, #state{route_mgr=RouteMgr, ipvs_mgr=IPVSMgr,
-        tree=Tree, nodes=Nodes, netns=Namespaces,
+-spec(handle_vips([{key(), [backend()]}], state()) -> state()).
+handle_vips(VIPs, #state{route_mgr=RouteMgr, ipvs_mgr=IPVSMgr,
```
GoelDeepak: Is there any specific reason for a different code structure in `handle_vips` and `handle_reconcile` ?

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -476,19 +498,19 @@ log_netns_diff(Namespaces, _PrevNamespaces) ->
 -spec(handle_netns_event(EventType, [netns()], state()) -> state()
     when EventType :: add_netns | remove_netns | reconcile_netns).
 handle_netns_event(remove_netns, ToDel,
-        #state{ipvs_mgr=IPVSMgr, route_mgr=RouteMgr, netns=Prev}=State) ->
+        #state{ipvs_mgr=IPVSMgr, route_mgr=RouteMgr, namespaces=Prev}=State) ->
     Namespaces = dcos_l4lb_route_mgr:remove_netns(RouteMgr, ToDel),
-    Namespaces = dcos_l4lb_route_mgr:remove_netns(IPVSMgr, ToDel),
+    Namespaces = dcos_l4lb_ipvs_mgr:remove_netns(IPVSMgr, ToDel),
```
GoelDeepak: 😱 This appeared to be a serious bug!!

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -44,6 +44,7 @@
 -type key() :: dcos_l4lb_mesos_poller:key().
 -type backend() :: dcos_l4lb_mesos_poller:backend().
 -type ipport() :: {inet:ip_address(), inet:port_number()}.
+-type namespace() :: term().
```
GoelDeepak: Any reason why we moved from `host | netns()` to a generic`term()`?

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
urbanserj: The logic is the same, but a bit different. In `handle_vips`, we apply all changes to all networking namespaces based on old known values. In `handle_reconcile` we check current kernel configuration and apply all changes to each networking namespaces separately.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
urbanserj: Yes, please see the comment message of the corresponding commit. There is a mess in types of network namespaces in dcos_l4lb. Some functions expect `netns()` (a.k.a. `#netns{}`), but some other requires `#netns.key`. There is no types in all other modules. So it is my mistake that I found out it while I was testing this patch. `dcos_l4lb_mgr` works with keys, not `netns()`. Key can be any `term()`.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
urbanserj: Most likely the process crashes in this case and supervisor restarts it. I think it's not a big deal.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
GoelDeepak: I understand that. My question was more on what we gain by having this difference?

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
GoelDeepak: wouldn't we miss removing IPVS configuration?

urbanserj: If I understand the code correctly, we remove network namespace from manager processes when it already not longer exists.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
urbanserj: Optimization? I don't see any other reasons.

PR #135, opened 2019-03-27
Add metrics for mesos listener

Fabs:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-4953

### DIFF
--- a/apps/dcos_l4lb/test/dcos_l4lb_mesos_poller_SUITE.erl
+++ b/apps/dcos_l4lb/test/dcos_l4lb_mesos_poller_SUITE.erl
@@ -29,7 +29,7 @@ end_per_suite(Config) ->
     Config.
 
 init_per_testcase(_, Config) ->
-    meck:new(dcos_net_mesos_listener, [no_link]),
+    meck:new(dcos_net_mesos_listener, [no_link, passthrough]),
     meck:expect(dcos_net_mesos_listener, poll, fun meck_mesos_poll/0),
 
     meck:new(dcos_l4lb_mgr, [no_link]),
--- a/apps/dcos_net/src/dcos_net_mesos_listener.erl
+++ b/apps/dcos_net/src/dcos_net_mesos_listener.erl
@@ -7,7 +7,8 @@
     is_leader/0,
     next/1,
     poll/0,
-    from_state/1
+    from_state/1,
+    init_metrics/0
 ]).
 
 %% gen_server callbacks
@@ -133,8 +134,10 @@ handle_info({'DOWN', _MonRef, process, Pid, _Info}, State) ->
 handle_info(_Info, State) ->
     {noreply, State}.
 
+terminate(normal, _State) ->
+    ok;
 terminate(_Reason, _State) ->
-    ok.
+    prometheus_counter:inc(mesos_listener, failures_total, [], 1).
 
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
@@ -852,22 +855,29 @@ handle_unsubscribe(Pid, #state{subs=Subs}=State) ->
 notify(_TaskId, _Task, #state{subs=undefined}=State) ->
     State;
 notify(TaskId, Task, #state{subs=Subs, timeout=Timeout}=State) ->
-    maps:fold(fun (Pid, Ref, ok) ->
-        Pid ! {task_updated, Ref, TaskId, Task},
-        ok
-    end, ok, Subs),
+    Begin = erlang:monotonic_time(),
+    try
+        maps:fold(fun (Pid, Ref, ok) ->
+            Pid ! {task_updated, Ref, TaskId, Task},
+            ok
+        end, ok, Subs),
 
-    maps:fold(fun (Pid, Ref, St) ->
-        receive
-            {next, Ref} ->
-                St;
-            {'DOWN', _MonRef, process, Pid, _Info} ->
-                handle_unsubscribe(Pid, St)
-        after Timeout div 3 ->
-            exit(Pid, {?MODULE, timeout}),
-            St
-        end
-    end, State, Subs).
+        maps:fold(fun (Pid, Ref, St) ->
+            receive
+                {next, Ref} ->
+                    St;
+                {'DOWN', _MonRef, process, Pid, _Info} ->
+                    handle_unsubscribe(Pid, St)
+            after Timeout div 3 ->
+                exit(Pid, {?MODULE, timeout}),
+                St
+            end
+        end, State, Subs)
+    after
+        prometheus_summary:observe(
+            mesos_listener, pubsub_duration_seconds,
+            [], erlang:monotonic_time() - Begin)
+    end.
 
 %%%===================================================================
 %%% Maps Functions
@@ -932,16 +942,19 @@ handle_init(State0) ->
             State0;
         {error, Error} ->
             lager:error("Couldn't connect to mesos: ~p", [Error]),
+            prometheus_counter:inc(mesos_listener, failures_total, [], 1),
             erlang:send_after(Timeout, self(), init),
             State0
     end.
 
 -spec(start_stream() -> {ok, state()} | {error, term()}).
 start_stream() ->
+    prometheus_boolean:set(mesos_listener, is_leader, [], false),
     Opts = [{stream, {self, once}}],
     Request = #{type => <<"SUBSCRIBE">>},
     case dcos_net_mesos:call(Request, [{timeout, infinity}], Opts) of
         {ok, Ref, Pid} ->
+            prometheus_boolean:set(mesos_listener, is_leader, [], true),
             httpc:stream_next(Pid),
             erlang:monitor(process, Pid),
             State = #state{pid=Pid, ref=Ref},
@@ -955,11 +968,14 @@ start_stream() ->
 -spec(handle_stream(binary(), state()) ->
     {noreply, state()} | {stop, term(), state()}).
 handle_stream(Data, State) ->
+    Size = byte_size(Data),
+    prometheus_counter:inc(mesos_listener, bytes_total, [], Size),
     case stream(Data, State) of
         {next, State0} ->
             {noreply, State0};
         {next, Obj, State0} ->
             State1 = handle(Obj, State0),
+            handle_metrics(State1),
             handle_stream(<<>>, State1);
         {error, Error} ->
             lager:error("Mesos protocol error: ~p", [Error]),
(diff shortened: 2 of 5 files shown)


Fabs:
Hey @urbanserj, it is not tested fully, but I would like your opinion if the metrics are being captured in the right places. So them tomorrow I can prepare a final version.

Fabs:
Also, I am unsure if I am initing the metrics in the right place.

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -997,3 +1011,55 @@ stream(Data, #state{pid=Pid, size=Size, buf=Buf}=State) ->
...
+        {labels, []},
+        {help, "Total number of bytes received"}]),
+    prometheus_counter:new([
+        {registry, mesos_listener},
+        {name, failures_total},
+        {labels, []},
+        {help, "Total number of bytes failures"}]),
+    prometheus_counter:new([
+        {registry, mesos_listener},
+        {name, redirects_total},
```
urbanserj: I think this metric should have a better name, a help string or/and even a type (i.e. boolean `standby`?)

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -974,9 +985,12 @@ stream(Data, #state{pid=Pid, size=undefined, buf=Buf}=State) ->
     case binary:split(Buf0, <<"\n">>) of
         [SizeBin, Tail] ->
             Size = binary_to_integer(SizeBin),
+            prometheus_counter:inc(mesos_listener, messages_total, [], 1),
+            prometheus_counter:inc(mesos_listener, bytes_total, [], Size),
             State0 = State#state{size=Size, buf= <<>>},
             stream(Tail, State0);
         [Buf0] when byte_size(Buf0) > 12 ->
+            prometheus_counter:inc(mesos_listener, failures_total, [], 1),
```
urbanserj: This one is trickier. Let me think how to implement it better, I guess we should increment it in `terminate/2` and when we receive an unexpected http response from mesos.

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -986,14 +1001,107 @@ stream(Data, #state{pid=Pid, size=Size, buf=Buf}=State) ->
...
+        {error, Error}
+    end.
+
+
+%%%===================================================================
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
```
Fabs: Had to break up this function due to size, but do not really like this approach. Do you have any alternatives?

apps/dcos_net/src/dcos_net_sup.erl
```diff
@@ -8,6 +8,11 @@ start_link() ->
     supervisor:start_link({local, ?MODULE}, ?MODULE, []).
 
 init([]) ->
+    try
```
Fabs: Here as well I need the `try` or some tests fail. I would rather have this initialization somewhere else where I would not need the try if possible. One other option is to move the try into the `init_metrics` function. What do you think?

apps/dcos_net/src/dcos_net_mesos_listener.erl
Fabs: There is something here I do not quite understand. Is the goal of the `redirects_total` metric to show in which moments of time are we being redirected? Do we plan to use it knowing that a redirect means that we are not talking to a master?

apps/dcos_net/src/dcos_net_mesos_listener.erl
Fabs: Do you have any new suggestion here?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -986,14 +1001,107 @@ stream(Data, #state{pid=Pid, size=Size, buf=Buf}=State) ->
...
+-spec(stream_decode(binary(), integer(), State) -> {error, term()} |
+     {next, State} | {next, jiffy:object(), State}
+         when State :: state()).
+stream_decode(Buf, Size, State) ->
+    <<Head:Size/binary, Tail/binary>> = Buf,
+    State0 = State#state{size=undefined, buf=Tail},
+    try
+        prometheus_counter:inc(mesos_listener, messages_total, [], 1),
+        % is the Size right here?
+        prometheus_counter:inc(mesos_listener, bytes_total, [], Size)
```
Fabs: You recommended me to move the `bytes_total` metric to the begining of `handle_stream`. It looks to me though that conceptually, everytime I increment `messages_total` by 1, I should also increment `bytes_total` by the Size of that specific message.

Do you see any issue with this approach?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -986,14 +1001,106 @@ stream(Data, #state{pid=Pid, size=Size, buf=Buf}=State) ->
...
+    end.
+
+
+%%%===================================================================
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_summary:new([
```
Fabs: @urbanserj what would trigger a subscriber to be added here? I have a minidcos cluster with one agent and one master. I have 0 subscribers. My cluster is running some apps, including one app with a VIP that keeps restarting so I could keep receiving status updates. But the `map:size(Subs)` is 0. 

I am looking to test if that metric is being reported properly.

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: Could you please break it up by domain area, not by metric type?

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: The first snapshot can take several seconds or even minutes. If the `bytes` metric is incremented here we don't not see that we are getting a huge state message from mesos.

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: You should have 1 subscriber on `leader.mesos`, it's `dcos_dns_mesos` [1].

[1] https://github.com/dcos/dcos-net/blob/master/apps/dcos_dns/src/dcos_dns_mesos.erl#L76

apps/dcos_net/src/dcos_net_sup.erl
urbanserj: Please let's work on the code and the logic. Once it works, we can start discussing the tests.

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: I'm sorry, I haven't had a chance to check it. I still think `terminate/2` and http errors are the best places for this metric. Could you please try to add a fake `handle_call` method that always stops the process with a non-`normal` reason? The metric must be incremented in this case.

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: Prometheus has native support for boolean values. it is supported in `prometheus.erl` library as well [1]. Just add it to the beginning of the list of all collectors [2].

[1] https://github.com/deadtrickster/prometheus.erl/blob/v4.2.2/src/metrics/prometheus_boolean.erl
[2] https://github.com/dcos/dcos-net/blob/b30888b/config/sys.config#L98

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -986,14 +1027,108 @@ stream(Data, #state{pid=Pid, size=Size, buf=Buf}=State) ->
...
+            maps:size(WT)),
+        prometheus_gauge:set(
+            mesos_listener, frameworks_total, [],
+            maps:size(F)),
+        prometheus_gauge:set(
+            mesos_listener, agents_total, [],
+            maps:size(A))
+    catch error:_Error ->
+        ok
+    end,
```
urbanserj: You actually don't need `try catch end` wrappers in this module for metrics. There is only one test that uses this module directly and it tests only this module. So it makes sense to fix the eunit test instead of adding wrappers here and, possibly, making the code less robust. It's a rare case when you have to use `try catch end` wrappers for metrics, e.g. a lot of tests that are calling the code, but not checking the results (dns zones) or other applications are incrementing counters (dns zones, logger metrics). It just works with a small patch:

```
diff --git a/apps/dcos_net/test/dcos_net_mesos_listener_tests.erl b/apps/dcos_net/test/dcos_net_mesos_listener_tests.erl
index 8ec6dd2..9e1fdb7 100644
--- a/apps/dcos_net/test/dcos_net_mesos_listener_tests.erl
+++ b/apps/dcos_net/test/dcos_net_mesos_listener_tests.erl
@@ -553,6 +553,9 @@ setup(FileName) ->
             Pid ! stream_next
         end),
 
+    ok = application:start(prometheus),
+    dcos_net_mesos_listener:init_metrics(),
+
     {ok, _Pid} = dcos_net_mesos_listener:start_link(),
     stream_wait(),
 
@@ -569,6 +572,8 @@ cleanup(_Tasks) ->
     exit(Pid, kill),
     exit(StreamPid, kill),
 
+    ok = application:stop(prometheus),
+
     meck:unload(httpc).
 
 stream_wait() ->
```

apps/dcos_net/src/dcos_net_sup.erl
urbanserj: Ok, it's a tricky one. Here is a crash report:

```
=CRASH REPORT==== 4-Apr-2019::21:09:24.361147 ===
  crasher:
    initial call: application_master:init/4
    pid: <0.4144.0>
    registered_name: []
    exception exit: {{undef,
                         [{dcos_net_mesos_listener,init_metrics,[],[]},
                          {dcos_net_sup,init,1,
                              [{file,
                                   "/home/circleci/dcos-net/apps/dcos_net/src/dcos_net_sup.erl"},
                               {line,12}]},
                          {supervisor,init,1,
                              [{file,"supervisor.erl"},{line,295}]},
                          {gen_server,init_it,2,
                              [{file,"gen_server.erl"},{line,374}]},
                          {gen_server,init_it,6,
                              [{file,"gen_server.erl"},{line,342}]},
                          {proc_lib,init_p_do_apply,3,
                              [{file,"proc_lib.erl"},{line,249}]}]},
                     {dcos_net_app,start,[normal,[]]}}
      in function  application_master:init/4 (application_master.erl, line 138)
    ancestors: [<0.4143.0>]
    message_queue_len: 1
    messages: [{'EXIT',<0.4145.0>,normal}]
```

it means there is no `dcos_net_mesos_listener:init_metrics/0`.

Let's add `false = dcos_net_mesos_listener:module_info()` to this test:

```
{badmatch,
     [{module,dcos_net_mesos_listener},
      {exports,[{module_info,0},{module_info,1},{poll,0}]},
      {attributes,
          [{vsn,[289324849627647645634493986770560104722]},
           {lager_records,
               [{state,
                    [pid,ref,size,buf,timeout,timeout_ref,agents,frameworks,
                     tasks,waiting_tasks,subs]}]}]},
      {compile,[{version,"7.3.2"},{options,[]}]},
      {native,false},
      {md5,<<217,169,240,108,123,178,51,84,132,179,28,36,148,194,57,18>>}]}
```

So only `dcos_net_mesos_listener:poll/0` function is exported. That's definitely not true if you check the code of the module. It means it's a fake module. There is a test that is mocking the module, but doesn't call the corresponding `meck:unmock/1` function.

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -986,14 +1027,108 @@ stream(Data, #state{pid=Pid, size=Size, buf=Buf}=State) ->
...
+    init_metrics_pubsub(),
+    init_metrics_mesos_state(),
+    init_metrics_received().
+
+init_metrics_pubsub() ->
+    prometheus_summary:new([
+        {registry, mesos_listener},
+        {name, pubsub_duration_seconds},
+        {labels, []},
+        {help, "Time spent processing pubsub"}]).
```
urbanserj: More context is needed here.

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -986,14 +1027,108 @@ stream(Data, #state{pid=Pid, size=Size, buf=Buf}=State) ->
...
+        {name, messages_total},
+        {labels, []},
+        {help, "Total number of messages"}]).
+
+init_metrics_mesos_state() ->
+    prometheus_gauge:new([
+        {registry, mesos_listener},
+        {name, agents_total},
+        {labels, []},
+        {help, "Total number of Agents"}]),
```
urbanserj: ditto: more context

ichernetsky: nit: why `Agents`, `Frameworks`, `Tasks` start with capital letters?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -942,11 +962,25 @@ start_stream() ->
...
+                    ok
+            end,
             httpc:stream_next(Pid),
             erlang:monitor(process, Pid),
             State = #state{pid=Pid, ref=Ref},
             {ok, handle_heartbeat(State)};
         {error, {http_status, {_HTTPVersion, 307, _StatusStr}, _Data}} ->
+            try
+                prometheus_boolean:set(
+                  mesos_listener, listening_leader_boolean,
```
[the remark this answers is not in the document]
Fabs: I was trying to fix the other indentation problem we usually have and I wrongly set some parameters. I fixed the indentation offset again to 4. I think this one we won't have again.

apps/dcos_net/src/dcos_net_mesos_listener.erl
Fabs: No good reason, fixed that.

apps/dcos_l4lb/test/dcos_l4lb_mesos_poller_SUITE.erl
```diff
@@ -31,6 +31,7 @@ end_per_suite(Config) ->
 init_per_testcase(_, Config) ->
     meck:new(dcos_net_mesos_listener, [no_link]),
     meck:expect(dcos_net_mesos_listener, poll, fun meck_mesos_poll/0),
+    meck:expect(dcos_net_mesos_listener, init_metrics, fun () -> ok end),
```
Fabs: When I looked at the issue of init, I found it was already doing `meck:unload(dcos_net_mesos_listener)`, so I decided to add a dummy init_metrics function to the mock. The tests pass :tada:

Fabs:
@urbanserj could you please look again? I think I did everything. I will still be here a few more hours tonight since I really wanna get this one merged this sprint.

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -942,11 +962,25 @@ start_stream() ->
...
+                  [], false)
```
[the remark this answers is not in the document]
urbanserj: @Fabs It has not been resolved.

apps/dcos_l4lb/test/dcos_l4lb_mesos_poller_SUITE.erl
urbanserj: It's a band-aid, not a proper fix. Could you please fix the root cause? `r3:break()` might help you [1]

[1] https://www.rebar3.org/v3/docs/using-breakpoints-to-debug-tests

urbanserj (approved):

Fabs:
@ichernetsky I heard you are great at checking the help strings. Would you mind taking a look?

ichernetsky (approved):
@Fabs, what you heard is simply wrong. I am exceptionally great at concurring with @urbanserj

PR #138, opened 2019-04-12
Conflict between VIP port and port mapping

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-4970
DC/OS PR: https://github.com/dcos/dcos/pull/5110
gen_netlink PR: https://github.com/mesosphere/gen_netlink/pull/20

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
@@ -1,0 +1,353 @@
+-module(dcos_l4lb_ipset_mgr).
+
+-include_lib("gen_netlink/include/netlink.hrl").
+
+-behaviour(gen_server).
+
+%% API
+-export([
+    start_link/0,
+    get_entries/0,
+    get_entries/1,
+    add_entries/2,
+    remove_entries/2,
+    cleanup/0
+]).
+
+%% gen_server callbacks
+-export([init/1, handle_continue/2,
+    handle_call/3, handle_cast/2]).
+
+-export_type([entry/0]).
+
+-type entry() :: dcos_l4lb_mesos_poller:key().
+
+-record(state, {
+    netlink :: pid()
+}).
+
+-define(IPSET_PROTOCOL, 6).
+-define(IPSET_TYPE, "hash:ip,port").
+-define(IPSET_NAME, "dcos-l4lb").
+-define(IPSET_NAME_IPV6, "dcos-l4lb-ipv6").
+
+
+-spec(get_entries() -> [entry()]).
+get_entries() ->
+    % NOTE: FOR DEBUG PURPOSE ONLY.
+    {ok, Pid} = start_link(),
+    try
+        get_entries(Pid)
+    after
+        erlang:unlink(Pid),
+        gen_server:cast(Pid, stop)
+    end.
+
+-spec(get_entries(pid()) -> [entry()]).
+get_entries(Pid) ->
+    gen_server:call(Pid, get_entries).
+
+-spec(add_entries(pid(), [entry()]) -> ok).
+add_entries(Pid, Entries) ->
+    gen_server:call(Pid, {add_entries, Entries}).
+
+-spec(remove_entries(pid(), [entry()]) -> ok).
+remove_entries(Pid, Entries) ->
+    gen_server:call(Pid, {remove_entries, Entries}).
+
+-spec(start_link() ->
+    {ok, Pid :: pid()} | ignore | {error, Reason :: term()}).
+start_link() ->
+    gen_server:start_link(?MODULE, [], []).
+
+%%%===================================================================
+%%% gen_server callbacks
+%%%===================================================================
+
+init([]) ->
+    {ok, {}, {continue, dcos_l4lb_config:ipset_enabled()}}.
+
+handle_continue(false, {}) ->
+    {noreply, disabled};
+handle_continue(true, {}) ->
+    {ok, Pid} = gen_netlink_client:start_link(?NETLINK_NETFILTER),
+    {ok, ?IPSET_PROTOCOL} = get_protocol_version(Pid),
+    {ok, []} = maybe_create_ipset(Pid, inet),
+    {ok, _} = maybe_add_iptables_rules(inet),
+    {ok, []} = maybe_create_ipset(Pid, inet6),
+    {ok, _} = maybe_add_iptables_rules(inet6),
+    {noreply, #state{netlink=Pid}}.
+
+handle_call(get_entries, _From, disabled) ->
+    {reply, [], disabled};
+handle_call(_Request, _From, disabled) ->
+    {reply, ok, disabled};
+handle_call(get_entries, _From, #state{netlink=Pid}=State) ->
+    {ok, EntriesIPv4} = get_entries(Pid, ?IPSET_NAME),
+    {ok, EntriesIPv6} = get_entries(Pid, ?IPSET_NAME_IPV6),
+    {reply, EntriesIPv4 ++ EntriesIPv6, State};
+handle_call({add_entries, Entries}, _From, #state{netlink=Pid}=State) ->
+    lists:foreach(fun ({Protocol, IP, Port}) ->
+        {ok, []} = add_entry(Pid, Protocol, IP, Port)
+    end, Entries),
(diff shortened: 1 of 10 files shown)


apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
```diff
@@ -0,0 +1,334 @@
...
+    {ok, []} | {error, atom() | non_neg_integer(), term()}).
+del_entry(Pid, Protocol, IP, Port) ->
+    cmd_entry(Pid, del, Protocol, IP, Port).
+
+-spec(cmd_entry(pid(), Command, Protocol, IP, Port) ->
+    {ok, []} | {error, atom() | non_neg_integer(), term()}
+        when Command :: add | del, Protocol :: protocol(),
+             IP :: inet:ip_address(), Port :: inet:port_number()).
+cmd_entry(Pid, Command, Protocol, IP, Port) ->
+    Name =
```
Fabs: Would that read better?
```erlang
Family = dcos_l4lb_app:family(IP),
Name = case Family of
    inet -> ?IPSET_NAME,
    inet6 -> ?IPSET_NAME_IPV6
end,
```

apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
```diff
@@ -0,0 +1,334 @@
...
+        4352 => ipset_err_type_specific
+    },
+    maps:get(Code, Mapping, Code).
+
+%%%===================================================================
+%%% Cleanup function
+%%%===================================================================
+
+-spec(cleanup() -> ok).
+cleanup() ->
```
Fabs: Should we use the `maybe_`  version of add here so when cleanup is performed we do not add rules if ipv6 is not enabled for instance?

apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
```diff
@@ -0,0 +1,334 @@
+-module(dcos_l4lb_ipset_mgr).
```
Fabs: Should we add a SUITE of tests for ipset at least to cover some basic cases? I think this could serve as a safety net for future changes, especially if done by other folks in the team.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -181,63 +184,111 @@ start_reconcile_timer() ->
 %%%===================================================================
 
 -spec(handle_reconcile([{key(), [backend()]}], state()) -> state()).
-handle_reconcile(VIPs, #state{route_mgr=RouteMgr, ipvs_mgr=IPVSMgr,
-        tree=Tree, nodes=Nodes, namespaces=Namespaces}=State) ->
+handle_reconcile(VIPs, #state{tree=Tree, nodes=Nodes, namespaces=Namespaces,
```
Fabs: Maybe also some more tests to cover that the ipset data was correctly changed during reconcile?

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -181,63 +184,111 @@ start_reconcile_timer() ->
...
+            DiffRoutes = dcos_net_utils:complement(Routes, PrevRoutes),
+
+            PrevVIPsP = get_vips(IPVSMgr, Namespace),
+            DiffVIPs = diff(PrevVIPsP, VIPsP),
 
-        PrevRoutes = get_routes(RouteMgr, Namespace),
-        {RoutesToAdd, RoutesToDel} =
-            dcos_net_utils:complement(Routes, PrevRoutes),
+            {Namespace, LogPrefix, DiffRoutes, DiffVIPs}
+        end, Namespaces),
```
Fabs: Have you thought about having metrics for ipset? If you have some, maybe we can add a ticket to add those metrics part of the l4lb epic?

Fabs: Oh. O just found that on the spreadsheet. Nice :).

apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
urbanserj: Good catch! I've fixed it.

apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
urbanserj: Unfortunately, there is no basic case here. The current set of tests are checking that ipset and new iptables rules are not breaking any basic cases. Also I added an integration test for this non-trivial case to dcos repo. Please see test_vip_port_mapping [1].

[1] https://github.com/dcos/dcos/pull/5110/files#diff-89643056102d7647724e7dc94be80843R278

apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
urbanserj: This function is used only in tests and will always be used in test only. We must clean all the things that this module creates. It doesn't matter if ipv6 is disabled (btw, it's always enabled in tests) or if any command fails.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
urbanserj: It's definitely out of scope for this ticket. It sounds like a good integration test rather than a unit test.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -181,63 +184,111 @@ start_reconcile_timer() ->
...
 
 -spec(handle_reconcile([{key(), [backend()]}], state()) -> state()).
-handle_reconcile(VIPs, #state{route_mgr=RouteMgr, ipvs_mgr=IPVSMgr,
-        tree=Tree, nodes=Nodes, namespaces=Namespaces}=State) ->
+handle_reconcile(VIPs, #state{tree=Tree, nodes=Nodes, namespaces=Namespaces,
+        route_mgr=RouteMgr, ipvs_mgr=IPVSMgr, ipset_mgr=IPSetMgr}=State) ->
     % If everything is ok this function is silent and changes nothing.
     VIPs0 = vips_port_mappings(VIPs),
     VIPs1 = healthy_vips(VIPs0, Nodes, Tree),
     VIPsP = prepare_vips(VIPs1),
```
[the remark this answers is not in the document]
urbanserj: `P` means `prepared`, but with `prepared` is a too long name (e.g. `PreparedPrevVIP`). `P` is a simplified version of `VIPs`. It has another data type with no agent ip address.

```
-spec(prepare_vips([{key(), [backend()]}]) -> [{key(), [ipport()]}]).
-type backend() :: {
    AgentIP :: inet:ip4_address(),
    {BackendIP :: inet:ip_address(), BackendPort :: inet:port_number() }
}.
-type ipport() :: {inet:ip_address(), inet:port_number()}.
```

apps/dcos_l4lb/src/dcos_l4lb_ipset_mgr.erl
```diff
@@ -0,0 +1,353 @@
...
+    {ok, Pid} = gen_netlink_client:start_link(?NETLINK_NETFILTER),
+    {ok, ?IPSET_PROTOCOL} = get_protocol_version(Pid),
+    {ok, []} = maybe_create_ipset(Pid, inet),
+    {ok, _} = maybe_add_iptables_rules(inet),
+    {ok, []} = maybe_create_ipset(Pid, inet6),
+    {ok, _} = maybe_add_iptables_rules(inet6),
+    {noreply, #state{netlink=Pid}}.
+
+handle_call(get_entries, _From, disabled) ->
+    {reply, [], disable};
```
urbanserj: JFI I've been testing disabled ipset and found a typo.

GoelDeepak: Thanks @urbanserj. Should have been caught in review 🤦‍♂️

urbanserj: No, it's my bad.

PR #150, opened 2019-05-08
Add Round-Robin DNS support

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-5118


apps/dcos_dns/src/dcos_dns_config.erl
```diff
@@ -100,3 +101,7 @@ mesos_resolvers() ->
 -spec(mesos_resolvers([upstream()]) -> ok).
 mesos_resolvers(Upstreams) ->
     application:set_env(?APP, mesos_resolvers, Upstreams).
+
+-spec(loadbalance() -> atom()).
+loadbalance() ->
+    application:get_env(?APP, loadbalance, round_robin).
```
GoelDeepak: instead of `loadbalance` shall we call it as `strategy`?

apps/dcos_dns/src/dcos_dns_handler.erl
```diff
@@ -361,6 +362,33 @@ max_payload_size(
...
+-spec(loadbalance(dns:message()) -> dns:message()).
+loadbalance(#dns_message{answers=Answers}=DNSMessage) ->
+    % NOTE: The order of CNAME records does matter.
+    {CNameRRs, RRs} =
+        lists:splitwith(
+            fun (#dns_rr{type = Type}) ->
+                Type =:= ?DNS_TYPE_CNAME
+            end, Answers),
+    Strategy = dcos_dns_config:loadbalance(),
+    RRs0 = loadbalance(Strategy, DNSMessage, RRs),
```
GoelDeepak: instead of `loadbalance` shall we call it `order`?

apps/dcos_dns/src/dcos_dns_config.erl
urbanserj: `loadbalance` is just how it is called in coredns [1], I'm fine with `strategy`.

[1] https://github.com/coredns/coredns/blob/4083852b70d4681752af8ad9a1840150fd52c071/plugin/loadbalance/README.md

apps/dcos_dns/src/dcos_dns_handler.erl
urbanserj: I've renamed it to `randomize`. Please let me know what you think.

PR #153, opened 2019-05-22
Codified networking run book

ichernetsky:
It's been implemented in collaboration with @jkoelker

### DIFF
--- a/runbook/README.md
+++ b/runbook/README.md
@@ -1,0 +1,40 @@
+# DC/OS Networking Run Book Automation
+
+This is a codified version of `dcos-net` part of the networking
+runbook.
+
+## Usage
+
+```
+$ net-runbook --help
+usage: net-runbook [options]
+
+a script for troubleshooting and collecting networking diagnostics data
+
+options:
+  --no-net-toolbox     do not use mesosphere/net-toolbox Docker image
+                       to capture some data, and use ipvsadm and such
+                       instead if they are available on the host
+  --help               print this message
+```
+
+It should be run under root. If possible the tool pulls
+`mesosphere/net-toolbox` Docker image that contains certain utilities
+that it uses to capture diagnostics data. If it fails to do so, it is
+expected that tools like `ipvsadm` and `ipset` are avaialbe on the
+host.
+
+After executing `net-runbook.sh`, there should be a tarball named like
+`net-runbook-<ip-address>.tar.gz` containing diagnostics data.
+
+## Recommendations
+
+What nodes to run the script on, depend on the type of networking
+issue being observed. By default it is good to run it on a master, an
+agent and a public agent. If there are multiple host networking
+subnets, then it is advised to have it run on at least one node from
+each subnet. If an issue is observed only on some nodes, the best
+approach is to run the script on those nodes and also on a node where
+the issue is not observed. In any case, if possible, it should be
+always executed on at least one master node, because some data is
+collected only on masters.
--- a/runbook/net-runbook.sh
+++ b/runbook/net-runbook.sh
@@ -1,0 +1,376 @@
+#!/bin/bash
+
+set -uo pipefail
+
+SCRIPT="$0"
+
+if [ -z "${DCOS_VERSION+x}" ]; then
+  exec /opt/mesosphere/bin/dcos-shell "$SCRIPT" "$@"
+fi
+
+usage() {
+  echo "usage: $SCRIPT [options]"
+  echo
+  echo "a script for troubleshooting and collecting networking diagnostics data"
+  echo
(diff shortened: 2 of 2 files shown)


runbook/runbook.sh
```diff
@@ -0,0 +1,298 @@
...
+if [ -z "${DCOS_VERSION+x}" ]; then
+  exec /opt/mesosphere/bin/dcos-shell "$SCRIPT"
+fi
+
+usage() {
+  echo "usage: $SCRIPT"
+  echo
+  echo "a networking helper for troubleshooting and collecting diagnostics data"
+  echo
+  echo "available commands:"
```
jkoelker: No subcommands are available (just data output).

urbanserj:
Hi @ichernetsky, thanks a lot! I do really appreciate your work.

At the moment I don't quite understand the purpose of the script and what the workflow will be. DC/OS Diagnostic Bundle [1] already contains a half of dns, all overlay, and almost all l4lb commands from the run book. From my experience, it allows to find root causes of 80% networking issues.

It would make sense if the script analyzes the data from diagnostic bundle and/or analyzes an issue on a cluster node in real time. Converting the run book to a script and collecting all the data would be a great first step towards that kind of script. Unfortunately, the script is written in bash and it will be almost impossible to do any analysis in the future.

[1] https://github.com/dcos/dcos/blob/9eee5cdb1fefbb9e64c630776baa383fdb53414e/gen/dcos-config.yaml#L1489-L1898

runbook/runbook.sh
ichernetsky: Thanks for catching this!

ichernetsky:
@urbanserj, thanks!

It is true that a lot of information that this script can capture, is already captured by the Diagnostics Bundle service, however this script makes it possible to collect more data on older versions of DC/OS, if something needs to be collected that is not collected already, it is easier to adjust this script and ask a customer to execute it on one or more nodes (in parallel), and this is just the beginning, I anticipate that more features will be added to this script in the future (let's say during next hackathon), such as testing connectivity, and not all features can be added to `dcos-config.yaml` I believe.

As for analyzing captured data, I would argue that it is a job for `bun` or even another tool that will be placed into this repo, that will analyze data collected by this script.

Does it make sense?

urbanserj:
@ichernetsky All supported versions (X-2 major versions and Y-4 minor versions) have almost all of the commands in the diagnostic bundle. Could you please elaborate on the workflow? Let's say someone hit a dns/l4lb issue in DC/OS 1.11.10. I'd just like to make sure that we don't ask to run the runbook script when it's not necessary (i.e. by default). That's my main concern here.

ichernetsky:
@urbanserj, in my opinion our version support policy is not defined by the arithmetic rule you mentioned as much as by particular versions that our biggest customer use, hence it is X-3 at the very least, and Y-Y for sure. Having said that, I might provide a hypothetical scenario in a hypothetical situation in which one would benefit from something that is not present in diagnostic bundle, when troubleshooting a DNS/L4LB problem in DC/OS of very specific version, but I think it is besides the point, because almost always it is possible to construct an example of something. I will later elaborate on why I believe there is a value in it. As for necessity to use it, I don't think a particular form of run book changes anything: it is not necessary to use the run book that is a Google doc, if you don't need it, and it is not necessary to use the run book that is a script. If you need it, you are welcome to benefit from it. The same way as the presence of diagnostics bundle service doesn't make it mandatory to collect a bundle every time a COPS ticket is created.

Here is my take on why it has a value:

- Essentially, it makes it possible to capture networking-related info anywhere.
- It is easy to extend the script in case something is not available in a bundle.
- It can be a foundation for future work in general and something like [mcli](https://github.com/timcharper/mcli) in particular. In addition to capturing information it can present bits of captured data in a digestible-to-human format. When there is such functionality, we can make this tool use the same directory/filename structures as the one used in the Diagnostics bundle service. This way we would be able to use this tool with diagnostics bundles too.
- It is another tool in a toolbox, which has its strengths.
- Quite often, diagnostics bundle collection fails on some nodes. In such a situation, it might be convenient to capture data on particular nodes using this script.

I could go on enumerating benefits, I see, but I think a sufficient number is given already. Please let me know if your concerns still remain.

ichernetsky:
@jkoelker @vinodkone could you please do another pass? Thank you!

ichernetsky:
@jkoelker thanks a bunch. I can't merge it yet, cause an approval is still required. One by someone from DC/OS networking team in Github. @GoelDeepak @urbanserj @orlandohohmeier could you please do me the favor?

ichernetsky:
Updated README with some recommendations on running the script.

PR #155, opened 2019-06-12
Use new pubsub interface from lashup_kv

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-5268
Lashup PR: https://github.com/dcos/lashup/pull/51

### DIFF
--- a/apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
+++ b/apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
@@ -1,218 +1,142 @@
-%%%-------------------------------------------------------------------
-%%% @author dgoel
-%%% @copyright (C) 2016, <COMPANY>
-%%% @doc
-%%%
-%%% @end
-%%% Created : 11 OCT 2016 9:27 PM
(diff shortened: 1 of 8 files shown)


apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
```diff
@@ -1,218 +1,127 @@
...
+%%% Internal functions
+%%%===================================================================
 
+-spec(wait_for_vtep() -> ok).
+wait_for_vtep() ->
+    % listener must wait until vtep is configured.
+    % dcos_overlay_poller polls local mesos agent module and
+    % configures vtep interfaces.
+    try
+        _Overlays = dcos_overlay_poller:overlays(),
```
GoelDeepak: what if `overlays` returns with an empty list? We would like to wait until we have a non-empty list of overlays.

urbanserj: Oops. Good catch! I didn't notice this case. Fixed.

PR #156, opened 2019-06-13
Force garbage collection

urbanserj:
JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-5269
Lashup PR: https://github.com/dcos/lashup/pull/52

### DIFF
--- a/apps/dcos_dns/src/dcos_dns_config_loader_server.erl
+++ b/apps/dcos_dns/src/dcos_dns_config_loader_server.erl
@@ -71,7 +71,7 @@ handle_info(?REFRESH_MESSAGE, State) ->
         error ->
             {ok, _} = timer:send_after(FailRefreshInterval, ?REFRESH_MESSAGE)
     end,
-    {noreply, State};
+    {noreply, State, hibernate};
 handle_info(Msg, State) ->
     lager:warning("Unhandled messages: ~p", [Msg]),
     {noreply, State}.
--- a/apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
@@ -95,7 +95,9 @@ handle_info(?GM_EVENTS(_R, _T)=Event, State) ->
 handle_info({lashup_kv_event, Ref, Key}, State) ->
     {noreply, handle_kv_event(Ref, Key, State)};
 handle_info({timeout, _Ref, reconcile}, State) ->
-    {noreply, handle_reconcile(State)};
+    State0 = handle_reconcile(State),
+    State1 = handle_gc(State0),
+    {noreply, State1, hibernate};
 handle_info(_Info, State) ->
     {noreply, State}.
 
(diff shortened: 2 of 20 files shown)


apps/dcos_dns/src/dcos_dns_config_loader_server.erl
```diff
@@ -71,7 +71,7 @@ handle_info(?REFRESH_MESSAGE, State) ->
         error ->
             {ok, _} = timer:send_after(FailRefreshInterval, ?REFRESH_MESSAGE)
     end,
-    {noreply, State};
+    {noreply, State, hibernate};
```
GoelDeepak: curious: why in certain cases we wait for a certain time before hibernating and why in others we immediately hibernate?

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -623,6 +625,19 @@ bes_port_mappings(PMs, Protocol, AgentIP, BEs) ->
                 {BEAgentIP, {BEIP, BEPort}}
         end, BEs).
 
+%%%===================================================================
+%%% GC funtions
+%%%===================================================================
+
+-spec(handle_gc(state()) -> state()).
+handle_gc(#state{ipvs_mgr=IPVSMgr, route_mgr=RouteMgr,
```
GoelDeepak: curious: why we have to explicitly garbage collect these structures instead of `hibernate`?

apps/dcos_dns/src/dcos_dns_config_loader_server.erl
urbanserj: I'm trying to avoid gc timer when it's possible, e.g. we call some function periodically, but not too frequently:

`dcos_dns_config_loader_server` sends a message to itself with a 2-minute delay, so it can call garbage collector after each call.

`dcos_dns_mesos` receives a lot of messages from `dcos_net_mesos_listener`, but pushes data to lashup only every n seconds, so it calls garbage collector only when pushes data to lashup. Also, it's just the best moment to call garbage collector because its state is significantly smaller after such push.

`dcos_dns_udp_server` receives udp packets and start new processes, it doesn't have periodic, but not too frequent events. In that case we have to add a gc timer.

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
urbanserj: It's the same reason as here: https://github.com/dcos/dcos-net/pull/156#discussion_r293197818. `dcos_l4lb_mgr` does reconciliation every 30 seconds, to avoid any other logic or gc timer, it explicitly garbage collect its child processes.

apps/dcos_dns/src/dcos_dns_config_loader_server.erl
GoelDeepak: Got it. Thanks

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
GoelDeepak: Thanks

PR #161, opened 2019-07-12
Overlay metrics

ichernetsky:
JIRA issue: [DCOS_OSS-5324](https://jira.mesosphere.com/browse/DCOS_OSS-5324)

Expose a bunch of overlay networking metrics.

In addition to the metrics, a couple of things are piggybacked:

- improved logging
- explicit error handling
- fixed indentation

### DIFF
--- a/apps/dcos_overlay/src/dcos_overlay_configure.erl
+++ b/apps/dcos_overlay/src/dcos_overlay_configure.erl
@@ -3,7 +3,7 @@
 -export([
     start_link/1,
     stop/1,
-    maybe_configure/2,
+    apply_config/2,
     configure_overlay/2
 ]).
 
@@ -12,8 +12,8 @@
 
 -spec(start_link(config()) -> pid()).
 start_link(Config) ->
-   Pid = self(),
-   proc_lib:spawn_link(?MODULE, maybe_configure, [Pid, Config]).
+    Pid = self(),
+    proc_lib:spawn_link(?MODULE, apply_config, [Pid, Config]).
 
 -spec(stop(pid()) -> ok).
 stop(Pid) ->
@@ -21,48 +21,111 @@ stop(Pid) ->
     exit(Pid, kill),
     ok.
 
--spec(maybe_configure(pid(), config()) -> term()).
-maybe_configure(Pid, Config) ->
-    {ok, Netlink} = dcos_overlay_netlink:start_link(),
-    try
-        lists:foreach(fun (Overlay) ->
-            maybe_configure_overlay(Netlink, Config, Overlay)
-        end, dcos_overlay_poller:overlays()),
-        Pid ! {dcos_overlay_configure, applied_config, Config}
-    after
-        dcos_overlay_netlink:stop(Netlink)
+-spec(apply_config(pid(), config()) -> ok | {error, term()}).
+apply_config(Pid, Config) ->
+    case apply_config(Config) of
+        ok ->
+            Pid ! {dcos_overlay_configure, applied_config, Config},
+            ok;
+        {error, Error} ->
+            Pid ! {dcos_overlay_configure, failed, Error, Config},
+            {error, Error}
     end.
 
--spec(maybe_configure_overlay(pid(), config(), jiffy:json_term()) -> ok).
+-spec(apply_config(config()) -> ok | {error, term()}).
+apply_config(Config) ->
+    case dcos_overlay_netlink:start_link() of
+        {ok, Netlink} ->
+            try maybe_configure(Netlink, Config) of
+                ok ->
+                    ok;
+                {error, Error} ->
+                    lager:error(
+                        "Failed to apply overlay config ~p due to ~p",
+                        [Config, Error]),
+                    {error, Error}
+            catch
+                Class:Error:StackTrace ->
+                    lager:error(
+                        "Failed to apply overlay config ~p due to ~p",
+                        [Config, {Class, Error, StackTrace}]),
+                    {error, {Class, Error}}
+            after
+                dcos_overlay_netlink:stop(Netlink)
+            end;
+        {error, Error} ->
+            lager:error(
+                "Failed to apply overlay config ~p due to ~p",
+                [Config, Error]),
+            {error, Error}
+    end.
+
+-spec(maybe_configure(pid(), config()) -> ok | {error, term()}).
+maybe_configure(Netlink, Config) ->
+    lists:foldl(
+        fun (Overlay, ok) ->
+                case maybe_configure_overlay(Netlink, Config, Overlay) of
+                    ok ->
+                        ok;
+                    {error, Error} ->
+                        lager:error(
+                            "Failed to configure overlay ~p due to ~p",
+                            [Overlay, Error]),
+                        {error, Error}
+                end;
+            (_, {error, Error}) ->
+                {error, Error}
+        end, ok, dcos_overlay_poller:overlays()).
+
+-spec(maybe_configure_overlay(pid(), config(), jiffy:json_term()) ->
+    ok | {error, term()}).
 maybe_configure_overlay(Pid, Config, #{<<"info">> := OverlayInfo} = Overlay) ->
     Subnet = maps:get(<<"subnet">>, OverlayInfo, undefined),
     #{<<"backend">> := #{<<"vxlan">> := VxLan}} = Overlay,
     #{<<"vtep_name">> := VTEPName} = VxLan,
-    ok = maybe_configure_overlay(Pid, VTEPName, Config, Subnet),
-    case application:get_env(dcos_overlay, enable_ipv6, true) of
-        true ->
-            Subnet6 = maps:get(<<"subnet6">>, OverlayInfo, undefined),
-            ok = maybe_configure_overlay(Pid, VTEPName, Config, Subnet6);
-        false ->
-            ok
+    case maybe_configure_overlay(Pid, VTEPName, Config, Subnet) of
+        ok ->
+            case application:get_env(dcos_overlay, enable_ipv6, true) of
+                true ->
+                    Subnet6 = maps:get(<<"subnet6">>, OverlayInfo, undefined),
+                    maybe_configure_overlay(Pid, VTEPName, Config, Subnet6);
+                false ->
+                    ok
+            end;
+        {error, Error} ->
+            {error, Error}
     end.
 
--spec(maybe_configure_overlay(pid(), binary(), config(), undefined) -> ok).
+-spec(maybe_configure_overlay(pid(), binary(), config(), undefined) ->
+    ok | {error, term()}).
 maybe_configure_overlay(_Pid, _VTEPName, _Config, undefined) ->
     ok;
 maybe_configure_overlay(Pid, VTEPName, Config, Subnet) ->
     ParsedSubnet = parse_subnet(Subnet),
     case maps:find(ParsedSubnet, Config) of
         {ok, OverlayConfig} ->
-            lists:foreach(fun ({VTEPIPPrefix, Data}) ->
-                maybe_configure_overlay_entry(
-                    Pid, VTEPName, VTEPIPPrefix, Data)
-            end, OverlayConfig);
+            maybe_configure_overlay_entries(Pid, VTEPName, OverlayConfig);
         error ->
             ok
     end.
 
--spec(maybe_configure_overlay_entry(pid(), binary(), subnet(), map()) -> ok).
+maybe_configure_overlay_entries(_Pid, _VTEPName, [] = _OverlayConfig) ->
+    ok;
+maybe_configure_overlay_entries(Pid, VTEPName, OverlayConfig0) ->
+    [{VTEPIPPrefix, Data} | OverlayConfig] = OverlayConfig0,
+    case maybe_configure_overlay_entry(Pid, VTEPName, VTEPIPPrefix, Data) of
+        ok ->
+            maybe_configure_overlay_entries(Pid, VTEPName, OverlayConfig);
+        {error, Error} ->
+            lager:error(
+                "Failed to configure overlay for VTEP IP "
+                "prefix: ~p; data: ~p; due to ~p",
+                [VTEPIPPrefix, Data, Error]),
+            {error, Error}
+    end.
+
+-spec(maybe_configure_overlay_entry(pid(), binary(), subnet(), map()) ->
+    ok | {error, term()}).
 maybe_configure_overlay_entry(Pid, VTEPName, VTEPIPPrefix, Data) ->
     LocalIP = dcos_overlay_poller:ip(),
     case maps:get(agent_ip, Data) of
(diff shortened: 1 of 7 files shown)


apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
```diff
@@ -136,3 +156,31 @@ to_str(Mac) when length(Mac) =:= 6 ->
...
+
+%%%===================================================================
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, lashup_events_received_total},
```
urbanserj: Let's not use `lashup` in metric names and help strings when it's possible

```suggestion
        {name, configuration_updates_total},
```

or maybe even better to add a summary metric for how much time it takes to process an event. it generate two time series, `num` (number of events) and `sum`. See `*_duration_seconds` in l4lb.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -84,18 +84,27 @@ handle_info(init, []) ->
...
     {noreply, #state{netlink = Pid}};
 handle_info(poll, State0) ->
-    State1 =
+    prometheus_counter:inc(overlay, mesos_state_polls_total, [], 1),
+    NewState =
         case dcos_net_mesos:poll("/overlay-agent/overlay") of
             {error, Reason} ->
-                lager:warning("Overlay Poller could not poll: ~p~n", [Reason]),
+                prometheus_counter:inc(
+                  overlay, mesos_state_poll_errors_total, [], 1),
```
urbanserj:
```suggestion
                    overlay, poll_errors_total, [], 1),
```

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -120,51 +129,108 @@ parse_response(State0 = #state{known_overlays = KnownOverlays}, AgentInfo) ->
...
     NewOverlays = Overlays -- KnownOverlays,
-    lists:foldl(fun add_overlay/2, State1, NewOverlays).
+    lists:foldl(
+      fun(Overlay, {ok, StateAcc}) ->
+          case add_overlay(Overlay, StateAcc) of
+              {ok, _} = Result ->
+                  prometheus_counter:inc(
+                    overlay, networks_added_total, [], 1),
+                  Result;
+              {error, ErrorMsg} = Error ->
```
urbanserj: May I kindly ask you to explicitly match and return tuples _everywhere_? It'd be much easier to follow snd understand the logic and returning types in that case. i.e.

```
   case foobar() of
       {ok, X} ->
           {ok, X};
       {error, Error} ->
           {error, Error}
    end.
```

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -120,51 +129,108 @@ parse_response(State0 = #state{known_overlays = KnownOverlays}, AgentInfo) ->
     State1 = State0#state{ip = IP1},
     Overlays = maps:get(<<"overlays">>, AgentInfo, []),
     NewOverlays = Overlays -- KnownOverlays,
-    lists:foldl(fun add_overlay/2, State1, NewOverlays).
+    lists:foldl(
+      fun(Overlay, {ok, StateAcc}) ->
+          case add_overlay(Overlay, StateAcc) of
+              {ok, _} = Result ->
+                  prometheus_counter:inc(
+                    overlay, networks_added_total, [], 1),
```
urbanserj:
```suggestion
                      overlay, networks_added_total, [], 1),
```

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -339,6 +493,34 @@ parse_vtep_mac(MAC) ->
         end,
         MACComponents).
 
+%%%===================================================================
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_counter:declare([
```
urbanserj: ditty: `new/1`

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -339,6 +493,34 @@ parse_vtep_mac(MAC) ->
...
+        {registry, overlay},
+        {name, mesos_state_polls_total},
+        {help, "Total number of times Mesos overlay state has beed polled."}]),
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, mesos_state_poll_errors_total},
+        {help, "Total number of errors while polling Mesos overlay state."}]),
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, networks_added_total},
```
urbanserj: Could you please add a gauge metric for a number of overlay networks instead of `networks_added_total`?

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -339,6 +493,34 @@ parse_vtep_mac(MAC) ->
...
+        {help, "Total number of overlay networks added."}]),
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, network_addition_errors_total},
+        {help, "Total number of errors while adding overlay networks."}]),
```
urbanserj: IIRC poller configures only `vtep` and doesn't add new overlay networks, so maybe `vtep_configuration_errors_total`?

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -339,6 +493,34 @@ parse_vtep_mac(MAC) ->
...
 
+%%%===================================================================
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, mesos_state_polls_total},
```
urbanserj: I think it's a misleading name, we don't poll mesos state here. I'd suggest `polls_total` or `local_polls_total` or a summary metric (e.g `poll_duration_seconds`; like there is a `call_duration_seconds` metric in `dcos_net_mesos` module for operatior api calls).

apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
```diff
@@ -136,3 +156,31 @@ to_str(Mac) when length(Mac) =:= 6 ->
...
+        {registry, overlay},
+        {name, configurations_applied_total},
+        {help, "Total number of configurations applied."}]),
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, configuration_failures_total},
+        {help, "Total number of overlay configuration failures."}]),
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, configurations_timed_out_total},
```
urbanserj: I'd merge it into `configuration_failures_total`, no need to differentiate these two cases in metrics.

apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
```diff
@@ -136,3 +156,31 @@ to_str(Mac) when length(Mac) =:= 6 ->
...
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, lashup_events_received_total},
+        {help, "Total number of Lashup events received."}]),
+    prometheus_counter:declare([
+        {registry, overlay},
+        {name, reconciliations_total},
+        {help, "Total number of overlay configuration reconciliations attempted."}]),
```
urbanserj: What kind of information can we get from this metric? It does reconciliation every 5 minutes anyway.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -110,61 +101,141 @@ code_change(_OldVsn, State, _Extra) ->
...
+    State1 = State0#state{ip = AgentIP},
     Overlays = maps:get(<<"overlays">>, AgentInfo, []),
     NewOverlays = Overlays -- KnownOverlays,
-    lists:foldl(fun add_overlay/2, State1, NewOverlays).
+    Results = [catch add_overlay(Pid, Overlay, AgentIP)
+               || Overlay <- NewOverlays],
+    {Successes, Failures} =
+        lists:partition(fun(ok) -> true; (_) -> false end, Results),
+    prometheus_gauge:set(overlay, networks_configured, [],
+        length(Successes) + length(KnownOverlays)),
```
urbanserj: Eventually, this metrics will be equal to `length(Overlays)` anyway cuz `known_overlays` contains all overlay networks. Which is fine, but code could be easier in that case.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -205,69 +278,138 @@ match(_Attr, _List) ->
...
-                [IfName, Error])
+            lager:error("Couldn't enable IPv6 on ~s interface due to ~p",
+                        [IfName, Error]),
+            {error, Error}
     end.
 
-maybe_add_ip_rule(#{<<"info">> := #{<<"subnet">> := Subnet},
-                    <<"backend">> := #{<<"vxlan">> := #{<<"vtep_name">> := VTEPName}}},
-                  #state{netlink = Pid}) ->
+maybe_add_ip_rule(Pid,
```
urbanserj:
```suggestion
maybe_add_ip_rule(Pid, Overlay) ->
    #{…} = Overlay,
```

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -339,6 +501,36 @@ parse_vtep_mac(MAC) ->
...
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_counter:new([
+        {registry, overlay},
+        {name, polls_total},
+        {help, "Total number of times local Mesos overlays have beed polled."}]),
+    prometheus_histogram:new([
```
urbanserj: ditto: summary

Actually, the poller calls `poll/1` function every 30-120 seconds, telegraf gets metrics from dcos-net every 30 seconds. so `sum` time series will have all information about each individual polling call.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -339,6 +501,36 @@ parse_vtep_mac(MAC) ->
         end,
         MACComponents).
 
+%%%===================================================================
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_counter:new([
```
urbanserj: ditto: we don't need it anymore, there's `poll_duration_seconds_cnt`

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -339,6 +501,36 @@ parse_vtep_mac(MAC) ->
...
+        {help, "Total number of errors while polling local Mesos overlays."}]),
+    prometheus_gauge:new([
+        {registry, overlay},
+        {name, networks_configured},
+        {help, "The number of overlay networks configured."}]),
+    prometheus_counter:new([
+        {registry, overlay},
+        {name, network_configuration_failures_total},
+        {help, "Total number of failures while configuring overlay networks."}]),
+    ok.
```
urbanserj: ditto: `s/,ok././`

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -339,6 +501,36 @@ parse_vtep_mac(MAC) ->
...
+        {registry, overlay},
+        {name, poll_errors_total},
+        {help, "Total number of errors while polling local Mesos overlays."}]),
+    prometheus_gauge:new([
+        {registry, overlay},
+        {name, networks_configured},
+        {help, "The number of overlay networks configured."}]),
+    prometheus_counter:new([
+        {registry, overlay},
+        {name, network_configuration_failures_total},
```
urbanserj: follow-up: https://github.com/dcos/dcos-net/pull/161#discussion_r304359807

I'd say we can just crash the process if there is a lashup error, lashup has its own metrics. Let's do that and rename the metric to `vtep_configuration_failures_total`?

apps/dcos_overlay/src/dcos_overlay_poller.erl
ichernetsky: This function clause relies on pattern matching the way it is done right now. Otherwise, the other clause should be removed, and `case` be introduced into the first clause.

apps/dcos_overlay/src/dcos_overlay_poller.erl
ichernetsky: Please refer to my response regarding `configuration_update_processing_duration_seconds_cnt`

apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
```diff
@@ -136,3 +171,29 @@ to_str(Mac) when length(Mac) =:= 6 ->
...
+        {registry, overlay},
+        {name, configuration_updates_total},
+        {help, "Total number of overlay configuration updates received."}]),
+    prometheus_histogram:new([
+        {registry, overlay},
+        {name, configuration_update_processing_duration_seconds},
+        {buckets, [0.01, 0.1, 1.0, 10.0, 60.0]},
+        {duration_unit, seconds},
+        {help, "The time spent processing overlay configuration updates."}]),
+    prometheus_counter:new([
```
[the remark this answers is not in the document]
urbanserj: It's true, but that's how Prometheus works. I'd say all prometheus queries are read-only and write-once anyway. The naming schema is super confusing to me. Nothing much we can do here, but only to follow the conversation and best practices. It is more weird when there are two time series with same data.

apps/dcos_overlay/src/dcos_overlay_poller.erl
urbanserj: Got it.

apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
```diff
@@ -136,3 +171,29 @@ to_str(Mac) when length(Mac) =:= 6 ->
...
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_counter:new([
+        {registry, overlay},
+        {name, configuration_updates_total},
+        {help, "Total number of overlay configuration updates received."}]),
+    prometheus_histogram:new([
+        {registry, overlay},
+        {name, configuration_update_processing_duration_seconds},
```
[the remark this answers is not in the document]
urbanserj: Can we remove `configuration_` from all metric names in this module? Does it add any information about the metrics?

apps/dcos_overlay/src/dcos_overlay_poller.erl
ichernetsky: If `eventually` is fine here, I am happy to adjust.

apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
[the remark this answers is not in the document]
ichernetsky: There is another subtle difference. `configuration_update_processing_duration_seconds` tracks duration of all updates, no matter what they result into (errors or not), whereas `configuration_updates_processed_total` tracks the number of updates that have been successfully applied.

apps/dcos_overlay/src/dcos_overlay_poller.erl
ichernetsky: I think that whether Lashup has its own metrics or not, should not affect what metrics are needed in this module, cause Lashup is a DB and all, whereas this module uses different terminology and operates with different beasts.

apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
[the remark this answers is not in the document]
urbanserj: That's not an issue at all. `update_processing_duration_seconds_cnt - update_failures_total` will give you the same results. We had exactly the same questions about summary/histogram and counter metrics in dns, l4lb, and lashup. It's kinda hard to get used to this mess 😞

apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
```diff
@@ -132,7 +164,27 @@ to_str({IP, Prefix}) ->
...
+
+%%%===================================================================
+%%% Metrics functions
+%%%===================================================================
+
+-spec(init_metrics() -> ok).
+init_metrics() ->
+    prometheus_counter:new([
+        {registry, overlay},
+        {name, updates_total},
```
urbanserj: there is `update_processing_duration_seconds_cnt`

apps/dcos_overlay/src/dcos_overlay_configure.erl
```diff
@@ -12,57 +12,122 @@
...
+        fun(Overlay, ok) ->
+            case maybe_configure_overlay(Netlink, Config, Overlay) of
+                ok ->
+                    ok;
+                {error, Error} ->
+                    lager:error("Failed to configure overlay ~p due to ~p",
+                        [Overlay, Error]),
+                    {error, Error}
+            end;
+           (_, {error, Error}) ->
```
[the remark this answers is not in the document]
ichernetsky: Hmm. I see 4- and 8-space indentation here. I guess it depends on from what angle you look at it. :)

apps/dcos_overlay/src/dcos_overlay_configure.erl
```diff
@@ -12,57 +12,122 @@
...
+
+-spec(apply_config(config()) -> ok | {error, term()}).
+apply_config(Config) ->
+    case dcos_overlay_netlink:start_link() of
+        {ok, Netlink} ->
+            try maybe_configure(Netlink, Config) of
+                ok ->
+                    ok;
+                {error, Error} ->
+                    lager:error("Failed to apply overlay config ~p due to ~p",
```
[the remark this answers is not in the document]
ichernetsky: Will fix.

apps/dcos_overlay/src/dcos_overlay_configure.erl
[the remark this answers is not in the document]
urbanserj: Your indentation is good, but you can do something like this only with emacs. Otherwise, you have to do 1-, 2-, 3-, 4-space indentation manually. We can't force everyone to use emacs, it's inhumane.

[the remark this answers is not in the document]
ichernetsky: I meant, that you use `(` as a reference in the example you provided. Whereas I used `fun` as a reference point, which is I believe consistent with indentation in case of named functions as opposed to anonymous functions in this particular case. But whatever, already adjusted. Consistency across the board is more important.

apps/dcos_overlay/src/dcos_overlay_configure.erl
```diff
@@ -12,57 +12,122 @@
...
-                maybe_configure_overlay_entry(
-                    Pid, VTEPName, VTEPIPPrefix, Data)
-            end, OverlayConfig);
+            maybe_configure_overlay_entries(Pid, VTEPName, OverlayConfig);
         error ->
             ok
     end.
 
--spec(maybe_configure_overlay_entry(pid(), binary(), subnet(), map()) -> ok).
+maybe_configure_overlay_entries(Pid, VTEPName, OverlayConfig) ->
```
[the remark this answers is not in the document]
ichernetsky: Awesome. Thanks. After working for a few months with Go, I need to get used to tail recursion again, it seems :)

apps/dcos_overlay/src/dcos_overlay_lashup_kv_listener.erl
ichernetsky: Hmm. There are updates from Lashup, which are accounted for by this metric, and reconciliations too, and for all of them `update_processing_duration_seconds` is used. Now I am thinking of a better word instead of `update`

ichernetsky:
Force-pushed. Waiting for a green light and then merging.

PR #169, opened 2019-08-10
add mesos basic auth

andreaspeters:
This PR will add the basic auth method to authenticate against the mesos master. Then the nice DC/OS net is also useful for "normal" mesos installations without authentication token. The authentication string have to be in a env variable and should looks like ```MESOS_BASIC_AUTH="Basic <BASE64STRING>"```.


ichernetsky:
@andreaspeters, thanks a lot for your contribution!

It is preferred not to supply individual configuration values through environment variables. Please refer to https://github.com/dcos/dcos-net/blob/master/apps/dcos_net/src/dcos_net_app.erl#L45 to learn more about the way `dcos-net` gets configured.

By default, `dcos-net` looks for configuration files under `/opt/mesosphere/etc/dcos-net.config.d`, which can be overwritten, for instance, through its application environment variable: `config_dir`. Please note that configuration files should be written in the Erlang term format.

andreaspeters:
Hi  @ichernetsky , thanks for your response. 
Actually, I'm not a Erlang dev, I just copy 'n paste what I already found in your code. And you already use a lot of  ```os:getenv("")```. :-) Well, for me it would not be easy to change it like u prefer. But now, you know and I think you understand why I wish me to have basic auth in dcos-net. If someone can add this feature (with a better peace of code), I would be very happy about it. :-)

PR #176, opened 2019-08-17
Leverage exponential backoff with requests to Mesos

ichernetsky:


apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
```diff
@@ -49,47 +49,64 @@ start_link() ->
...
 %%%===================================================================
 
--spec(start_poll_timer() -> reference()).
-start_poll_timer() ->
-    Timeout = dcos_l4lb_config:agent_poll_interval(),
-    erlang:start_timer(Timeout, self(), poll).
-
--spec(handle_poll() -> ok).
-handle_poll() ->
+-spec(handle_poll(backoff:backoff()) -> {ok, backoff:backoff()}).
```
urbanserj: there's no need to return a tuple with `ok` here, it's the only option.

apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
```diff
@@ -49,47 +49,64 @@ start_link() ->
...
+handle_call(Request, _From, State) ->
+    lager:warning("Unexpected request: ~p", [Request]),
     {reply, ok, State}.
 
-handle_cast(_Request, State) ->
+handle_cast(Request, State) ->
+    lager:warning("Unexpected request: ~p", [Request]),
+    {noreply, State}.
+
+handle_info({timeout, _TimerRef, poll}, State) ->
```
urbanserj: Could you please store a reference in process state and explicitly match `TimerRef` as it was before your changes?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -142,6 +153,22 @@ terminate(_Reason, _State) ->
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+reset_state(State) ->
+    case State#state.ref of
```
urbanserj: May I kindly ask you to not use `Foo#bar.baz` in dcos-net code? It's not readable. We prefer explicit matches, i.e. `#bar{baz = Baz} = Foo`.

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -142,6 +153,22 @@ terminate(_Reason, _State) ->
...
+        undefined -> ok;
+        Ref -> ok = httpc:cancel_request(Ref)
+    end,
+    erlang:cancel_timer(State#state.timeout_ref),
+    erlang:cancel_timer(State#state.backoff_timer_ref),
+    State#state{pid = undefined, ref = undefined, size = undefined,
```
urbanserj: That kind of reset can cause some issues in the future when a new field will be added to the state, but will not be reset. Can we return just a new `#state{}` here?

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -74,39 +83,23 @@ handle_call(Request, _From, State) ->
...
-                        lager:error(
-                            "Failed to parse overlay response due to ~p",
-                            [Error]),
-                        exit(Error)
-                end
-        end,
-    timer:send_after(State#state.poll_period, poll),
-    {noreply, State, hibernate};
-handle_info(_Info, State) ->
+handle_info({timeout, _TimerRef0, poll}, State) ->
```
urbanserj: ditto: `TimerRef`

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -142,6 +153,23 @@ terminate(_Reason, _State) ->
...
+%%% Internal functions
+%%%===================================================================
+
+reset_state(State) ->
+    #state{ref = Ref} = State,
+    case Ref of
+        undefined -> ok;
+        _ -> ok = httpc:cancel_request(Ref)
+    end,
+    erlang:cancel_timer(State#state.timeout_ref),
```
urbanserj: ditto, i.e.

```
reset_state(#state{ref = Ref, timerout_ref = TimeoutRef, backoff_timer_ref = …} = State) ->
```

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: In that case, can we explicitly set only subscribers in a new `#state{}`? `cancel_timer(FooRef), #state{subs=Subs}.`. All of those default values look not ok here, especially constants like `15000`.

ichernetsky:
@urbanserj, done. Please take another look.

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -117,10 +111,43 @@ code_change(_OldVsn, State, _Extra) ->
...
+    SuccessBackoff = backoff:init(
+        ?MIN_SUCCESS_POLL_PERIOD, ?MAX_SUCCESS_POLL_PERIOD, self(), poll),
+    FailureBackoff0 = backoff:init(
+        ?MIN_FAILURE_POLL_PERIOD, ?MAX_FAILURE_POLL_PERIOD, self(), poll),
+    FailureBackoff = backoff:type(FailureBackoff0, jitter),
+    #state{success_backoff = SuccessBackoff,
+        failure_backoff = FailureBackoff, netlink = Pid}.
+
+handle_poll(#state{netlink = Pid, known_overlays = Overlays,
+    success_backoff = SuccessBackoff0,
```
urbanserj: nit, ditto: one more indentation

apps/dcos_overlay/src/dcos_overlay_poller.erl
```diff
@@ -26,8 +27,16 @@
...
--define(MIN_POLL_PERIOD, 30000). %% 30 secs
--define(MAX_POLL_PERIOD, 120000). %% 120 secs
+
+%% A failed request is retried sooner than a successful one.
+%% The rationale is, the overlay configuration doesn't change
+%% frequently and the cost of applying such a configuration
+%% far outweighs the cost of fetching it from Mesos over HTTP.
+-define(MIN_SUCCESS_POLL_PERIOD, 30000).  %% 30  secs
+-define(MAX_SUCCESS_POLL_PERIOD, 120000). %% 120 secs
+-define(MIN_FAILURE_POLL_PERIOD, 1000).   %% 1   sec
```
urbanserj: I think 1 second is too frequently. To update overlay configuration you should re-configure the whole dc/os cluster and restart all dc/os systemd units. We don't really need backoff here, but if we have it, it shouldn't poll mesos overlay agent module more often than 30 seconds in any cases. One configuration for success and failure case is fine. Here is what I saw once I stopped mesos slave:

```
18:10:11.301 [warning] Overlay Poller could not poll: {failed_connect,[{to_address,{"172.17.0.4",5051}},{inet,[inet],econnrefused}]}
18:10:12.305 [warning] Overlay Poller could not poll: {failed_connect,[{to_address,{"172.17.0.4",5051}},{inet,[inet],econnrefused}]}
18:10:13.357 [warning] Overlay Poller could not poll: {failed_connect,[{to_address,{"172.17.0.4",5051}},{inet,[inet],econnrefused}]}
18:10:15.207 [warning] Overlay Poller could not poll: {failed_connect,[{to_address,{"172.17.0.4",5051}},{inet,[inet],econnrefused}]}
18:10:19.537 [warning] Overlay Poller could not poll: {failed_connect,[{to_address,{"172.17.0.4",5051}},{inet,[inet],econnrefused}]}
```

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -142,6 +153,20 @@ terminate(_Reason, _State) ->
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+reset_state(#state{ref = Ref, timeout_ref = TRef, subs = Subscribers,
+    backoff = Backoff, backoff_timer_ref = BackoffTRef}) ->
```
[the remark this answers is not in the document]
ichernetsky: Will do. Though at this point I can't help but wonder if it would be easier just to use `erlang-mode` or some other editor mode as a reference implementation instead of circling back and forth in PRs bringing the code into alignment with the undocumented set of style rules. What do you think? Does it make sense to document the guidelines for all the future references?

apps/dcos_overlay/src/dcos_overlay_poller.erl
ichernetsky: Exactly. This was one of the reasons why I brought it up a few days back in the channel. And there are two options that make sense in my opinion:

1. Not backing off at all. Why backing off up to 2 minutes, and not to `+infinity`, if changes to overlays rarely happen if ever? Especially, if `dcos-net` would be restarted anyways in case of such a change. It would be simpler just to poll the endpoint frequently enough (but not too frequently) and calculate a hash of Mesos response, and apply a configuration if its hash is different.
2. Since there was a exponential backoff in the first place for the success case, it follows that there should be one for the failure case too, and its minimal delay should be smaller, because fetching is cheap. Basically, what's stated in the comment fully applies here.

The way I see it, one configuration is self-contradicting which makes you wonder just like it made me wonder when I read it for the first time.

So what do we choose?

cc @GoelDeepak

ichernetsky:
@urbanserj, I failed to reproduce this one, but found another one, while trying. I completely forgot to notify subscribers about the end of stream. I think this is the reason behind the erroneous behavior you encountered. Thanks!

apps/dcos_overlay/src/dcos_overlay_poller.erl
GoelDeepak: I vote for option 1

urbanserj: it makes sense to poll overlay module on start more often cuz it may take some time to get the configuration from master overlay module. I'd just poll it every 30 seconds with no backoff and  no caching/hashing logic at all.

apps/dcos_net/src/dcos_net_mesos_listener.erl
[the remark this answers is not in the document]
urbanserj: I totally understand your frustration. The code style standard or document makes sense. However, I'm not sure if it worth the effort (we can discuss the reasons in person). I would love to have a fmt tool, but it doesn't exist. `erlang-mode` is a kinda standard that a very few projects forces and it works only if emacs is the only editor which is not [our] case. Also we have several code styles in one project, some of them are very bad. I'd recommend to use the same code style that is used in the module where you are making changes and also check the style in any big modules that have been added in last couple of years.

apps/dcos_overlay/src/dcos_overlay_poller.erl
ichernetsky: As far as I understand, there was a (performance-related) reason to introduce an exponential backoff retries for successful requests back then. Something has changed that renders unneeded anymore? Are you pro not doing hash calculation in order to periodically ensure that the configuration is actually applied?

urbanserj: We don't really care about performance in the poller, it polls a few tiny json objects.

GoelDeepak: > As far as I understand, there was a (performance-related) reason to introduce an exponential backoff retries for successful requests back then. Something has changed that renders unneeded anymore? Are you pro not doing hash calculation in order to periodically ensure that the configuration is actually applied?

There were two concerns when we introduced this functionality back then:
1. we wanted to converge faster in case of failure
2. Reduce unnecessary cpu cycles that in applying the same configuration over and over again. 

We cannot use a hash-based approach because we have seen instances where some other application would change the configuration unhood which would break overlay. So, the trick is to keep applying the config after a fixed interval.

ichernetsky: Should I revert this file then?

GoelDeepak: yeah, should be fine.

ichernetsky:
Ready for another pass, guys.

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -102,40 +103,49 @@ from_state(Data) ->
...
-    {ok, []}.
+    Interval = application:get_env(dcos_net, mesos_reconnect_timeout, 2000),
+    MaxInterval = application:get_env(
+        dcos_net, mesos_reconnect_max_timeout, 30000),
+    Backoff0 = backoff:init(Interval, MaxInterval, self(), init),
+    Backoff = backoff:type(Backoff0, jitter),
+    {ok, #state{backoff = Backoff}, {continue, init}}.
 
 handle_call(is_leader, _From, State) ->
     % There is no state if it's not connected
```
urbanserj: @ichernetsky it seems like now dcos_net_mesos_listener always thinks it's the leader, even if it runs on non-leader master nodes

ichernetsky: Nice catch. A fix is coming.

ichernetsky: BTW, is an immediate DC/OS bump with the fix warranted?

urbanserj: it only affects the mesos dns exporter, `dcos_dns_mesos_dns`. So I guess we can wait for the PR that fixes partial mesos state.

PR #179, opened 2019-08-27
Wait till agents become active before fanning out Mesos tasks

ichernetsky:
JIRA issue: [DCOS_OSS-5463](https://jira.mesosphere.com/browse/DCOS_OSS-5463)


apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -888,49 +932,78 @@ subscribe(Pid) ->
...
+-spec(pubsub(term(), state()) -> state()).
+pubsub(Message, #state{subs = Subscribers} = State) ->
+    pubsub(Message, Subscribers, State).
+
+-spec(pubsub(term(), #{pid() => reference()}, state()) -> state()).
+pubsub(_Message, _Subscribers, #state{ready = false} = State) ->
+    State;
+pubsub(Message, Subscribers, #state{timeout = Timeout} = State0) ->
     Begin = erlang:monotonic_time(),
     try
```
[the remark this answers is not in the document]
ichernetsky: You mean, why this function is changed? To accommodate the fact, that the listener sends a several types of messages to its subscribers. Hence there is this generic one that expects message acknowledgements for all the message types. Did it make it clear? What makes it harder to read it for you?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -888,49 +932,78 @@ subscribe(Pid) ->
...
+            pubsub_tasks(#{Pid => Ref}, State)
     end.
 
 -spec(handle_unsubscribe(pid(), state()) -> state()).
 handle_unsubscribe(Pid, #state{subs = Subs} = State) ->
     State#state{subs = maps:remove(Pid, Subs)}.
 
--spec(notify(task_id(), task(), state()) -> state()).
-notify(TaskId, Task, #state{subs = Subs, timeout = Timeout} = State) ->
+-spec(pubsub(term(), state()) -> state()).
```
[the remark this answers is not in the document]
ichernetsky: I can assure you I have no hard feelings even towards this suggestion of me having hard feelings towards a mere word.

For now there is a need to send messages of multiple types (tasks, task updates and so forth), `notify_*` function names would be unnecessary long. For instance, `notify_about_task_update`. Initially, I thought of `publish` as a common prefix, but then decided to keep it aligned with that was already used in a name of one of the metrics: `pubsub` as in pub/sub messaging. I can change it to `publish` if you like it more.

ichernetsky:
@urbanserj, sure, I will see what the best option is out there.

ichernetsky:
@urbanserj, after a bit of prototyping and consideration, I don't think that `#state.waiting_tasks` should come into play here.

1. In the typical scenarios, all data is received as a `get_state` response and through the consequent updates without any (logical) reordering of them.
2. Even in case of `#state.waiting_tasks` is empty, it doesn't mean that there is no need to wait for some time more. This is the case when events are received in the logical order. For extreme instance, the listener receives all agent and framework updates before receiving even a single task update.

Hence to keep as simple as possible, I am inclined to take the route of simply adding another timeout to wait for tasks.

ichernetsky:
@urbanserj it is ready for another pass. Thanks.

ichernetsky:
@urbanserj @GoelDeepak the state of listener grew. Do you think the complexity is still manageable? Is it worth introducing a middle-man that would be a front man to subscribers, and its job would be to manage subscribers, accumulate tasks, receive updates from the listeners, and send updates to subscribers?

apps/dcos_net/src/dcos_net_mesos_listener.erl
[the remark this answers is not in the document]
urbanserj: yeah, could you please change it to `publish`?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -925,13 +960,88 @@ notify(TaskId, Task, #state{subs = Subs, timeout = Timeout} = State) ->
...
+    Timeout = application:get_env(dcos_net,
+        mesos_agents_readiness_timeout, timer:minutes(10)),
+    erlang:start_timer(Timeout, self(), await_agents).
+
+-spec(start_tasks_readiness_timer() -> reference()).
+start_tasks_readiness_timer() ->
+    Timeout = application:get_env(dcos_net,
+        mesos_tasks_readiness_timeout, timer:minutes(1)),
+    erlang:start_timer(Timeout, self(), await_tasks).
+
```
urbanserj: nit: could you please add `%%%=\n%%% Phase functions\n%%%=` comment here?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -925,13 +960,88 @@ notify(TaskId, Task, #state{subs = Subs, timeout = Timeout} = State) ->
...
+maybe_next_phase(#state{phase = init} = State, update) ->
+    maybe_next_phase_init(State, update);
+maybe_next_phase(#state{phase = await_agents} = State, Reason) ->
+    maybe_next_phase_await_agents(State, Reason);
+maybe_next_phase(#state{phase = await_tasks} = State, Reason) ->
+    maybe_next_phase_await_tasks(State, Reason).
+
+-spec(maybe_next_phase_init(state(), update) -> state()).
+maybe_next_phase_init(#state{recovered_agents = RA} = State, update) ->
+    case map_size(RA) of
```
urbanserj: nit: would it make sense to use a function guard here and below, i.e

```
foobar(#{recovered_agents = RA} = State)
        when map_size(RA) > 0 ->
   …;
foobar(State) ->
   ….
```

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -925,13 +960,88 @@ notify(TaskId, Task, #state{subs = Subs, timeout = Timeout} = State) ->
...
+
+-spec(maybe_next_phase_await_agents(state(), timeout | update) -> state()).
+maybe_next_phase_await_agents(State, timeout) ->
+    TRef = start_tasks_readiness_timer(),
+    State#state{phase = await_tasks, recovered_agents = #{},
+        await_agents_timer_ref = make_ref(),
+        await_tasks_timer_ref = TRef};
+maybe_next_phase_await_agents(#state{recovered_agents = RA,
+        await_agents_timer_ref = AwaitAgentsTRef} = State, update) ->
+    case map_size(RA) of
```
urbanserj: nit: ditto, guard(?)

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -925,13 +960,88 @@ notify(TaskId, Task, #state{subs = Subs, timeout = Timeout} = State) ->
...
+        _ ->
+            State
     end.
 
+-spec(maybe_next_phase_await_tasks(state(), timeout | update) -> state()).
+maybe_next_phase_await_tasks(State, update) ->
+    State;
+maybe_next_phase_await_tasks(#state{await_tasks_timer_ref = TRef} = State,
+        timeout) ->
+    _ = erlang:cancel_timer(TRef),
```
urbanserj: nit: that's not necessary, it already timed out at that moment

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -49,12 +49,16 @@
     timeout = 15000 :: timeout(),
     timeout_ref = make_ref() :: reference(),
```
urbanserj: nit: you added few more timers to the state and this one is confusing, may I kindly ask you to rename it to `http_timeout` and `http_timeout_ref`?

apps/dcos_net/src/dcos_net_mesos_listener.erl
```diff
@@ -925,13 +960,88 @@ notify(TaskId, Task, #state{subs = Subs, timeout = Timeout} = State) ->
...
+-spec(start_agents_readiness_timer() -> reference()).
+start_agents_readiness_timer() ->
+    Timeout = application:get_env(dcos_net,
+        mesos_agents_readiness_timeout, timer:minutes(10)),
+    erlang:start_timer(Timeout, self(), await_agents).
+
+-spec(start_tasks_readiness_timer() -> reference()).
+start_tasks_readiness_timer() ->
+    Timeout = application:get_env(dcos_net,
+        mesos_tasks_readiness_timeout, timer:minutes(1)),
```
urbanserj: Ok, if we just wait for one more minute to get all the tasks, I think we are fine with non-empty `waiting_tasks` and unknown frameworks. Basically, we utilize one timeout for both, tasks and frameworks. Could you please check with the mesos if it's possible to safely use shorter timeout here, e.g. 10/30 seconds?

apps/dcos_net/test/dcos_net_mesos_listener_tests.erl
```diff
@@ -533,11 +565,21 @@ hello_overlay_setup() ->
...
+setenv() ->
+    application:set_env(dcos_net, is_master, true),
+    application:set_env(dcos_net, mesos_agents_readiness_timeout,
+        timer:seconds(1)),
+    application:set_env(dcos_net, mesos_tasks_readiness_timeout,
+        timer:seconds(1)).
+
 setup(FileName) ->
     Lines = read_lines(FileName),
+    setenv(),
```
urbanserj: If the application is not loaded `application:set_env/n` functions will do nothing. I guess there is a side effect and the application is already loaded by some other tests. Could you please set configuration after `application:load(dcos_net)`?

apps/dcos_net/src/dcos_net_mesos_listener.erl
urbanserj: I have one more question here: is it possible to check how long mesos is running, so if dcos-net / mesos listener restarted and there are no `recovered_agents` in `GET_STATE` then we can safely skip this timeout?

ichernetsky: Interesting idea. Yes, it is possible. There is the `GET_MASTER` call:

```
  // Contains the master's information.
  message GetMaster {
    optional MasterInfo master_info = 1;

    optional double start_time = 2;

    optional double elected_time = 3;
  }
```

Do you want me to implement it in this PR or a separate one?

urbanserj: It'd be a bit sad if we have to add one more call for that. Could you please check if it is not possible to re-use agents/frameworks `(re-)registered_time` from `GET_STATE`?

urbanserj: Once the timeout has been changed to 10 seconds, I think it's not a big deal. If it's easy to implement, let's do it. Otherwise, it doesn't worth the effort and I'm fine with the timeout.

PR #182, opened 2019-10-25
DC/OS overlay networks should be compared by-value

ichernetsky:
When deciding whether to push an overlay network update to Lashup or not, only VTEP IP address and subnet were used previously, which led to dropped updates, whereas all the other fields like MAC address should be used too.

JIRA issue: [DCOS_OSS-5620](https://jira.mesosphere.com/browse/DCOS_OSS-5620)


ichernetsky:
It pushes an update to Lashup only after it has been applied locally (meaning the interface, routes and all have been successfully configured). This PR just makes sure that none of such updates is dropped.

To solve the problem of configuring overlays on a node in the presence of conflicts after overlay state wipe out, we need to think of some sort of reconciliation: check what configuration is on the node, and wipe it out if it differs from the most up-to-date one, then apply the latter.

PR #190, opened 2020-01-14
Save netstat output

ichernetsky:
Also,
 - make the script compatible with DC/OS >= 2.0
 - fix indentation
 - make it use jq only if it is available

### DIFF
--- a/runbook/net-runbook.sh
+++ b/runbook/net-runbook.sh
@@ -1,3 +1,4 @@
+# -*- mode: shell-script; sh-basic-offset: 2 -*-
 #!/bin/bash
 
 set -uo pipefail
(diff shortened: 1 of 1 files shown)


runbook/net-runbook.sh
```diff
@@ -91,13 +97,21 @@ wrap-ipvsadm() {
...
 }
 
+maybe-pprint-json() {
+  if hash jq 2> /dev/null; then
+    jq .
+  else
+    cat
+  fi
+}
+
```
greggomann: Could you split the changes into this PR into separate commits? Since many of the changes here are cleanup, it would be nice to at least separate the netstat changes, the conditional pretty printing, and other cleanup.

orlandohohmeier:
@greggomann I agree that we should split these changes into different commits for future maintenance, but don't want to hold the PR any longer.

PR #191, opened 2020-01-30
Make runbook a bit more sysadmin friendly

jkoelker:
I've been using the runbook to automate the gathering of data while troubleshooting various issues and reproducing test failures. I've updated it accordingly to:

1. Pass [`shellcheck`](https://github.com/koalaman/shellcheck) linting
1. Use `dig` from the `net-toolbox` docker image
1. Wrap the main logic into a function that gets called at the end to ensure it is executed atomiclly.


jkoelker:
> What about we later remove `no-net-toolbox` option and handle the if-else logic internally in this script by first trying commands on the host and then docker container.

I like that idea, I'll add it to my todo list!

PR #193, opened 2020-02-20
Handle TASK_KILLING state for loadbalancing

jkoelker:
Set the weight to `0` for tasks in state `TASK_KILLING`. This allows existing connections to gracefully terminate while not accepting new connections to a backend.

JIRA: https://jira.d2iq.com/browse/COPS-5602
DCOS PR: https://github.com/dcos/dcos/pull/6907

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
@@ -118,35 +118,32 @@ handle_poll(true, Backoff0) ->
 
 -spec(handle_poll_state(#{task_id() => task()}) -> ok).
 handle_poll_state(Tasks) ->
-    HealthyTasks = maps:filter(fun is_healthy/2, Tasks),
+    RunningTasks = maps:filter(fun is_running/2, Tasks),
 
-    PortMappings = collect_port_mappings(HealthyTasks),
+    PortMappings = collect_port_mappings(RunningTasks),
     dcos_l4lb_mgr:local_port_mappings(PortMappings),
 
-    VIPs = collect_vips(HealthyTasks),
+    VIPs = collect_vips(RunningTasks),
     ok = push_vips(VIPs),
 
     LocalBackends = lists:sum([length(BEs) || BEs <- maps:values(VIPs)]),
     prometheus_gauge:set(l4lb, local_tasks, [], maps:size(Tasks)),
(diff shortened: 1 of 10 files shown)


jkoelker:
Not sure why circleci isn't reporting back:

https://circleci.com/gh/dcos/dcos-net/1446

apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
```diff
@@ -39,7 +39,9 @@
 -type lkey() :: {key(), riak_dt_orswot}.
 -type backend() :: {
     AgentIP :: inet:ip4_address(),
-    {BackendIP :: inet:ip_address(), BackendPort :: inet:port_number() }
+    {BackendIP :: inet:ip_address(),
+     BackendPort :: inet:port_number(),
+     BackendWeight :: inet:port_number() }
```
mainred: Maybe `non_neg_integer()` is more suitable that `inet:port_number()` for weight type in this case?

jkoelker: The max value for weight is 65535 (16bit integer). I just used `inet:port_number` so not to have to write a type just for this.

orlandohohmeier: You should be able to just use `non_neg_integer` here - no need to write a new type for this one.

jkoelker: @orlandohohmeier but that would allow invalid larger values than 65535, which are invalid weights. Whereas [`inet:port_number`](https://github.com/erlang/otp/blob/c5cb5ff89b5667b2e17946ed54b5f053a48e55c5/lib/kernel/src/inet.erl#L103) is already restricted appropriately.

jkoelker:
I'm going to move this back to WIP for the time being, the last commit is a significant behaviour change (but one for the better) and I don't feel comfortable getting it merged the day before I go on holiday ;)

apps/dcos_l4lb/src/dcos_l4lb_mesos_poller.erl
orlandohohmeier: gotcha, I'd vote for a new type then.

mainred: Thanks for your explanation.

orlandohohmeier:
Fixed the project configuration and Circle CI should be able to report back again.

PR #197, opened 2020-04-02
Vip weight regression

jkoelker:
JIRA: https://jira.d2iq.com/browse/COPS-6002
DCOS PR: https://github.com/dcos/dcos/pull/7186

### DIFF
--- a/apps/dcos_l4lb/src/dcos_l4lb_ipvs_mgr.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_ipvs_mgr.erl
@@ -282,7 +282,7 @@ handle_add_dest(ServiceIP, ServicePort, DestIP, DestPort, Protocol,
 handle_add_dest(Pid, Service, IP, Port, Family, Weight) ->
     Base = [{fwd_method, ?IP_VS_CONN_F_MASQ}, {weight, Weight}, {u_threshold, 0}, {l_threshold, 0}],
     Dest = [{port, Port}] ++ Base ++ ip_to_address(IP),
-    lager:info("Adding backend ~p to service ~p~n", [{IP, Port}, Service]),
+    lager:info("Adding backend ~p to service ~p~n", [{IP, Port, Weight}, Service]),
     Msg = #new_dest{request = [{dest, Dest}, {service, Service}]},
     case gen_netlink_client:request(Pid, Family, ipvs, [], Msg) of
         {ok, _} -> ok;
--- a/apps/dcos_l4lb/src/dcos_l4lb_ipvs_mgr.erl
+++ b/apps/dcos_l4lb/src/dcos_l4lb_ipvs_mgr.erl
@@ -16,6 +16,7 @@
 
 -export([get_dests/3,
          add_dest/8,
+         mod_dest/8,
          remove_dest/7,
          get_services/2,
          add_service/5,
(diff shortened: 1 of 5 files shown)


apps/dcos_l4lb/src/dcos_l4lb_ipvs_mgr.erl
```diff
@@ -268,6 +279,24 @@ handle_get_dests(Service, Namespace, #state{netns = NetnsMap, family = Family})
...
+                      Weight :: backend_weight()) -> ok | error).
+handle_mod_dest(ServiceIP, ServicePort, DestIP, DestPort, Protocol,
+                Namespace, #state{netns = NetnsMap, family = Family}, Weight) ->
+    Pid = maps:get(Namespace, NetnsMap),
+    Protocol1 = netlink_codec:protocol_to_int(Protocol),
+    Service = ip_to_address(ServiceIP) ++ [{port, ServicePort}, {protocol, Protocol1}],
+    Base = [{fwd_method, ?IP_VS_CONN_F_MASQ}, {weight, Weight}, {u_threshold, 0}, {l_threshold, 0}],
+    Dest = [{port, DestPort}] ++ Base ++ ip_to_address(DestIP),
+    lager:info("Modifying backend ~p for service ~p~n", [{DestIP, DestPort, Weight}, Service]),
+    Msg = #set_dest{request = [{dest, Dest}, {service, Service}]},
```
wavesoft, line 294: Damn, it took me a while to find the `set_dest` record 😄 I didn't knew we had forked the `gen_netlink` module...

apps/dcos_l4lb/src/dcos_l4lb_mgr.erl
```diff
@@ -669,6 +676,20 @@ local_port_mappings() ->
 
 -ifdef(TEST).
 
+-define(BE4, {{1, 2, 3, 4}, 80, 1}).
+-define(BE4OLD, {{1, 2, 3, 5}, 80}).
+
+be_port_mapping_test() ->
```
wavesoft, line 682: 👍

jkoelker:
> It looks good to me, but I haven't seen the part that sets the weight to 0. Is this coming from another component?

So for this branch (1.13.x) it was already merged in https://github.com/dcos/dcos-net/pull/194 and for master in https://github.com/dcos/dcos-net/pull/193. It never made it into 2.0.x, I just reused that PR to also add the fix ;)

apps/dcos_l4lb/src/dcos_l4lb_ipvs_mgr.erl
jkoelker: Yea :(, I wish we had renamed it after it looked like upstream wasn't going to maintain it.

apps/dcos_l4lb/src/dcos_l4lb_lashup_vip_listener.erl
```diff
@@ -120,11 +124,34 @@ process_vip(Key, BEs) ->
 
 -spec(categorize_backends([backend()]) -> [{family(), [backend()]}]).
 categorize_backends(BEs) ->
-    lists:foldl(fun ({_AgentIP, {IP, _Port, _Weight}}=BE, Acc) ->
+    lists:foldl(fun (BE, Acc) ->
+        case BE of
+            {_AgentIP, {IP, _Port, _Weight}} ->
+                BE0 = BE;
+            {AgentIP, {IP, Port}} ->
+                BE0 = {AgentIP, {IP, Port, 1}}
```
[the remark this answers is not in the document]
jkoelker: Correct, this handles the case were we get a backend that isn't weighted.

1.  In the 2nd case AgentIP is used since we want it to be bound on the match for that case. For the 1st case using the `_` prefix denotes a match, but unused binding, there's a linter check that checks for unused bindings.
2. Yea the `_Weight` binding was added in #194 since that is what changed the format of the `backend` type.

PR #200, opened 2020-08-28
runbook: capture more data

jkoelker:
Since the network runbook can be used as a lite "bundle" capture more
data.


jongiddy (approved):
:+1:

PR #201, opened 2020-09-01
runbook: capture vxlan state and nclock

jkoelker:


runbook/net-runbook.sh
```diff
@@ -322,17 +322,23 @@ overlay-data() {
...
+  bridge fdb > "$DATA_DIR/bridge-fdb.txt"
   ifconfig -a > "$DATA_DIR/ifconfig.txt"
-  ip link > "$DATA_DIR/ip-link.txt"
   ip addr > "$DATA_DIR/ip-addr.txt"
-  ip route > "$DATA_DIR/ip-route.txt"
+  ip link > "$DATA_DIR/ip-link.txt"
   ip neigh > "$DATA_DIR/ip-neigh.txt"
   ip ntable > "$DATA_DIR/ip-ntable.txt"
+  ip route > "$DATA_DIR/ip-route.txt"
+  ip route show table 42 > "$DATA_DIR/ip-route-42.txt"
```
jongiddy, line 332: Will table 42 always exist? Does it matter if it does not and this fails? We're not using `set -e` so I expect we just keep going if there are problems...

jongiddy (approved):
Looks good. It appears that if a command fails in this script, we keep going, which is probably what we want.

runbook/net-runbook.sh
jkoelker: It *should* always exist as that's how the overlay routing gets injected. I will continue if it fails, but that's a good call to handle / capture it so its in the runbook correctly instead of just a blank file if it does fail. I added capturing `stderr` as well. Thanks!

